PR #84, opened 2018-10-15
echo-server with secio and remote key extraction

echupriyanov:
* make secio mandatory
* extract remote public key for future use

### DIFF
--- a/examples/echo-server.rs
+++ b/examples/echo-server.rs
@@ -67,11 +67,6 @@ fn main() {
                println!("Remote addr: {:?}", addr);
                out.stream
              })
-            // upgrade::map(secio,
-            //  |out: SecioOutput<_>| {
-            //    println!("Remote key: {:?}", out.remote_key);
-            //    out.stream
-            //  })
         })
 
         // On top of plaintext or secio, we will use the multiplex protocol.
--- a/examples/floodsub.rs
+++ b/examples/floodsub.rs
@@ -30,10 +30,11 @@ extern crate tokio_stdin;
 use futures::Stream;
 use futures::future::Future;
 use std::{env, mem};
-use libp2p::core::{either::EitherOutput, upgrade};
+use libp2p::core::upgrade;
 use libp2p::core::{Multiaddr, Transport, PublicKey};
 use libp2p::peerstore::PeerId;
 use libp2p::tcp::TcpConfig;
+use libp2p::secio::SecioOutput;
 use libp2p::websocket::WsConfig;
 
 fn main() {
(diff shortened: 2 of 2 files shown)


examples/echo-server.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // Determine which address to listen to.
+    let listen_addr = env::args()
+        .nth(1)
+        .unwrap_or("/ip4/0.0.0.0/tcp/10333".to_owned());
+
+    // We start by creating a `TcpConfig` that indicates that we want TCP/IP.
+    let transport = TcpConfig::new()
+        // In addition to TCP/IP, we also want to support the Websockets protocol on top of TCP/IP.
+        // The parameter passed to `WsConfig::new()` must be an implementation of `Transport` to be
```
rtsisyk: Why do we need WebSockets on the top of TCP/IP. Can we use plain TCP/IP without web-sockets? I think that websockets are almost dead and superseded by HTTP/2.0. IT would be better to avoid using websockets for the new projects.

examples/echo-server.rs
```diff
@@ -0,0 +1,149 @@
...
+        // used for the underlying multiaddress.
+        .or_transport(WsConfig::new(TcpConfig::new()))
+
+        // On top of TCP/IP, we will use either the plaintext protocol or the secio protocol,
+        // depending on which one the remote supports.
+        .with_upgrade({
+            let secio = {
+                let private_key = include_bytes!("test-rsa-private-key.pk8");
```
rtsisyk: `Config` module can be used instead of hard-coded pathes.

examples/echo-server.rs
```diff
@@ -0,0 +1,149 @@
...
+            })
+        },
+    );
+
+    // We now use the controller to listen on the address.
+    let address = swarm_controller
+        .listen_on(listen_addr.parse().expect("invalid multiaddr"))
+        // If the multiaddr protocol exists but is not supported, then we get an error containing
+        // the original multiaddress.
+        .expect("unsupported multiaddr");
```
rtsisyk: Use `Result<(), Error>` instead of except/panic.

examples/echo-server.rs
```diff
@@ -0,0 +1,149 @@
...
+    // The address we actually listen on can be different from the address that was passed to
+    // the `listen_on` function. For example if you pass `/ip4/0.0.0.0/tcp/0`, then the port `0`
+    // will be replaced with the actual port.
+    println!("Now listening on {:?}", address);
```
rtsisyk: We have a logger now.

examples/echo-server.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    // `swarm_future` is a future that contains all the behaviour that we want, but nothing has
+    // actually started yet. Because we created the `TcpConfig` with tokio, we need to run the
+    // future through the tokio core.
+    tokio_current_thread::block_on_all(swarm_future.for_each(|_| Ok(()))).unwrap();
```
rtsisyk: ```swarm_future.for_each(|_| Ok(()))``` looks very strange for me. I guess that this line starts event loop in the current thread.

Cargo.toml
```diff
@@ -33,16 +33,32 @@ stegos_runtime = { path = "./runtime" }
...
 slog = "~2.4"
+rand = "0.5.5"
 slog-async = "~2.3"
 slog-term = "~2.4"
+tokio-codec = "0.1.1"
+tokio-io = "0.1.9"
+tokio-stdin = "0.1.1"
+tokio-current-thread = "0.1.3"
+
+libp2p = { git = "https://github.com/libp2p/rust-libp2p.git", rev = "e179951c74cb0c8916d27329f2dfd78917dddfb2" }
```
rtsisyk: why it is needed to restrict libp2p version by git hash?

Cargo.toml
```diff
@@ -33,16 +33,32 @@ stegos_runtime = { path = "./runtime" }
...
 
+[[example]]
+name = "floodsub"
+
+[[example]]
+name = "echo-server"
+
 [dev-dependencies]
 rust-libpbc = { git = "https://github.com/stegos/rust-pbcintf.git" }
+env_logger = "0.5.13"
```
rtsisyk: Please use `slog` module which we already have integrated.

crypto/src/pbc/mod.rs
```diff
@@ -23,6 +23,8 @@
 // SOFTWARE.
 
 #![allow(non_snake_case)]
+#![allow(dead_code)]
```
rtsisyk: Unrelated.

network/src/lib.rs
```diff
@@ -6,7 +6,15 @@ use slog::Logger;
...
+  
+}
+
 pub fn init(cfg: ConfigNetwork, log: Logger) -> Result<(), Box<Error>> {
     info!(log, "starting network"; "strval" => &cfg.strval, "u32val" => cfg.u32val);
     Ok(())
 }
+
+pub fn server() {
+  println!("Main loop");
```
rtsisyk: Please go ahead and integrate echo-server to the main application.

PR #89, opened 2018-10-17
Multithreaded FloodSub example

echupriyanov:
Obsoletes #84     

Separated network part and main loop

    * floodsub node runs in swarm in separate thread
    * messages are sent and received through separate async channels
    * requests to connect are also sent through async channel.


dbmcclain:
I fixed up the conflicts in the Cargo.toml and Cargo.lock files

PR #91, opened 2018-10-17
Implement Rust-style hashing using SHA3-256

### LINKED ISSUE #26: Choose Hashing Algo

rtsisyk:
Follow std::hash approach for hashing.
Now it possible to define hashing logging for any struct:

    // Define some structure
    struct Data {
        a: String,
        b: u8,
        c: u32,
    }

    // Define custom hashing logging
    impl Hashable for Data {
        fn hash(&self, state: &mut Hasher) {
            self.a.hash(state);
            self.b.hash(state);
            self.c.hash(state);
        }
    }

    // Calculate the hash
    let mut hasher = Hasher::new();
    d.hash(&mut hasher);
    let h2: Hash = hasher.result();

My initial version re-used std::hash::Hash and std::hash::Hasher.
Unfortunally, this approach has failed completely. Rust's
std::hash has some special logic for handling built-in types.
For example, for String they call hash(str.as_bytes()) + hash(255).
It is probably needed to workaround hashing of empty strings.
The same problem happens with arrays. I want the clear result
without extra garbage in the hash function.

Closes #26

### DIFF
--- a/crypto/src/hash/mod.rs
+++ b/crypto/src/hash/mod.rs
@@ -24,6 +24,9 @@
 // SOFTWARE.
 
 use sha3::{Digest, Sha3_256};
+use std::fmt;
+use std::mem;
+use std::slice;
 use utils::*;
 
 // -----------------------------------------------------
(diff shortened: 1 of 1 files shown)


dbmcclain:
hash of empty strings is correct, and I show an example of doing that in one of my test routines. I wonder if the hash(255) is a tag?

PR #99, opened 2018-10-22
Dbm working 5

dbmcclain:
work in progress - make visible with PR. Not asking for review just yet...


dbmcclain:
Ready to commit tentative sources...

Bulletproof construction - gives other enough information to others, to construct Bulletproof data items, while I work on finishing up the BP Validation code. Runs about the same speed as our previous Lisp/C version. 57 ms / BP.

PR #102, opened 2018-10-24
Proof-of-concept transport muxing

echupriyanov:
* implemented echo-reply protocol `rust-libp2p` way, compatible with
  MT Tokio Runtime (echo-server example)
* handle finished connections in `floodsub` example
* `muxedsub` example serving both echo and floodsub on the same ip/port


echupriyanov:
Obsoleted by #103

PR #105, opened 2018-10-26
Implement prunable Merkle Tree

### LINKED ISSUE #35: Merkle Trees

rtsisyk:
Closes #35

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1880,6 +1880,7 @@ name = "stegos_blockchain"
 version = "0.1.0"
 dependencies = [
  "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


rtsisyk:
> And so, e.g., if a [u8;32] has the same memory layout as a [u64;4] they will hash to the same value. But if the first were declared as a Lev32 and the second as a U256 in my code, then I will hash the first with a "Lev32" prefix, and the second with a "U256" prefix, to make certain that they hash to different values.
>
>This is certainly not mandatory at the moment, but I follow this practice for the same reasons that I form null-structs like struct Lev32([u8;32]), and struct U256([u64;4]). It serves as defensive programming to avert later errors that might arise from un-tagged data.
>
>My own experience shows that this is good practice. And our previous Lisp effort was a disaster at one point because people sloppily used un-tagged data, and I had to finally force everyone to use tagged data items. Let's hope between our new team and the Rust compiler, that the same situation can be avoided going forward.

It makes sense. I will follow your approach.

PR #109, opened 2018-10-31
Initial PeerStore implementation

echupriyanov:
* Initialize and populate store for known peers
* Use configured and detected IP addesses known addresses for out PeerId

Issue #85, opened 2018-10-15
Add PeerStore to store peer ids received.


echupriyanov:
Store remote peer's keys (ids) in shared PeerStore for future re-use.

echupriyanov:
Closed by #112

PR #112, opened 2018-10-31
Initial PeerStore and directed communications implementation

### LINKED ISSUE #111: Simulate direct communications using floodsub

echupriyanov:
* Initialize and populate store for known peers
* Use configured and detected IP addesses known addresses for out PeerId
* Simulate sending messages to particular peer, using floodsub and dedicated topics

Obsoletes #109 
Closes #111 #85


echupriyanov:
@dbmcclain 
To make things easier right now, I've implemented 2 methods on `node` object:
1. `node.set_id(id)` - sets this node identifier (we can use hash of public key here)
1. `node.send_to_id(id, Vec<u8>)` - sends direct message to node, identified by `id`

Right now its implemented as sending broadcast message to topic == node_id with single subscriber.

When `rust-libp2p` stabilizes enough, I'll reimplement these methods as real direct communications.

But as now, this should keep as going

PR #113, opened 2018-10-31
Refactor genesis block and EncryptedPayload

rtsisyk:
- Add hardcoded genesis block
- Switch to real crypto primitives for EncryptedPayload
- Fix Hash::from_str() to decode Hash from HEX.

See #36


crypto/src/hash/mod.rs
```diff
@@ -54,11 +55,10 @@ impl Hash {
         hasher.result()
     }
 
-    pub fn from_str(msg: &str) -> Hash {
-        // produce a Hash from a single string
-        let mut hasher = Hasher::new();
-        (*msg).hash(&mut hasher);
-        hasher.result()
+    pub fn from_str(s: &str) -> Result<Hash, hex::FromHexError> {
```
dbmcclain: This doesn't look correct to me... you are passing off the byte string as its own hash value. That certainly isn't correct for any of the crypto primitives that depend on this primitive routine... Actually, it would be a complete disaster because then everyone could know the discrete log of the hash points on Elliptic curves. Big no no. Makes it trivial to forge signatures if that ever happens.

dbmcclain: No... I actually need the security of SHA3-256 here.

crypto/src/pbc/secure.rs
```diff
@@ -213,6 +213,13 @@ impl SecretKey {
     pub fn to_str(&self) -> String {
         u8v_to_typed_str("SKey", &self.base_vector())
     }
+
```
dbmcclain: okay... I see you are just refactoring code here. That's fine

crypto/src/pbc/secure.rs
```diff
@@ -240,6 +247,11 @@ impl PublicKey {
     pub fn to_str(&self) -> String {
         u8v_to_typed_str("PKey", &self.base_vector())
     }
+
```
dbmcclain: hmmm.... probably not a good idea to refactor in this way... To be sure, you need to check on the expected lengths of the G1, G2, and Zr fields. 

Unless they are all identical, this will be trouble...

dbmcclain: ... and if they are identical, be sure to leave a breadcrumb comment about this shortcut. It will need to be reviewed if we ever (likely) change to new curves.

crypto/src/hash/mod.rs
rtsisyk: This function was needed to create valid Hash() objects from hex strings generated by SHA3-256 algorithm. It's just an inverse function for to_str(). I see that almost all primitives in stegos_crypto have to_str()/from_str(). Moreover, Hash also provides from_vector() which does the same job but for byte arrays. Of course, I didn't plan to use this constructor to create flawed sha3-256 hashes.

Anyway, if you think that we must never need to create Hash() object from it's hex representation, I can remove it and replace by Hasher::digest("string") in my code.

crypto/src/pbc/secure.rs
rtsisyk: Yes, initial from_str() was used to calculate SHA3-256 from the source string. I found that from_str() usually means "from hex" and decided to change this function to meet this convention.

crypto/src/pbc/secure.rs
rtsisyk: I added an assertion that the total length of `payload` inside EncryptedPayload::new() is 60 bytes. It should be enough.

crypto/src/hash/mod.rs
rtsisyk: https://github.com/stegos/stegos/pull/113/files#diff-5b2f2e83aba2947db9da40c54ab784b6R43
The only case I needed this constructor. Shall we remove it and replace with byte vector or Hasher() call?

blockchain/src/payload.rs
```diff
@@ -51,18 +55,32 @@ impl fmt::Display for EncryptedPayload {
...
-    pub fn garbage() -> EncryptedPayload {
+    pub fn new_payment(delta: Zr, gamma: Fr, amount: i64, pkey: PublicKey) -> EncryptedPayload {
+        // Convert amount to BE vector.
+        use std::mem::transmute;
+        let amount_bytes: [u8; 8] = unsafe { transmute(amount.to_be()) };
+        let gamma_bytes: [u8; 32] = gamma.bits().to_lev_u8();
+
+        let payload: Vec<u8> = [&amount_bytes[..], delta.base_vector(), &gamma_bytes[..]].concat();
+        // Ensure that the total length of package is 64.
+        assert_eq!(PAYLOAD_PAYMENT_ID.len() + payload.len(), 64);
```
rtsisyk: I check the total length of gamma and delta here. I found that ibe_encrypt() prepends `id` to the `payload`, therefor the total length is payload.len() + id.len()

crypto/src/pbc/secure.rs
rtsisyk: See https://github.com/stegos/stegos/pull/113/files#diff-b232303f528d490806a38972e103ee13R76

crypto/src/hash/mod.rs
dbmcclain: Heh! While producing the checksum test on the crypto libs, I ran into the same need. So now I understand what you were doing. I created a function Hash::basic_from_str() that takes a hex string representation of a Hash, and gives a byte vector Hash object as you and I both need.

The existing Hash::from_str() continues to work as before, and serves to hash strings with a real SHA3-256 hash. That's how it is needed in the crypto libs.

rtsisyk:
Re-based and pushed with David's changes. Let's open a ticket for all follow-ups.

rtsisyk:
I'll open a new pull request.

PR #127, opened 2018-11-05
clean up PRNG usage.

dbmcclain:
All modules now use the ThreadRng thread-local PRNG, which is (according to docs) a crypto-safe PRNG, seeded from entropy

create function Hash::from_hash_facsimile_str() to allow importing hash values from constant strings
Hash::from_str() is solely for the purpose of hashing arbitrary strings, akin to Hash::from_vector() for u8 vectors.
get things running unit tests properly again...

### DIFF
--- a/crypto/src/hash/mod.rs
+++ b/crypto/src/hash/mod.rs
@@ -47,7 +47,8 @@ impl Hash {
         self.0
     }
 
-    pub fn basic_from_str(hexstr: &str) -> Result<Self, hex::FromHexError> {
+    pub fn from_hash_facsimile_str(hexstr: &str) -> Result<Self, hex::FromHexError> {
+        // use this function to import a Hash digest facsimile from a string constant
         let mut v = [0u8; HASH_SIZE];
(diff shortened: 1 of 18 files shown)


crypto/src/hash/mod.rs
```diff
@@ -47,7 +47,8 @@ impl Hash {
         self.0
     }
 
-    pub fn basic_from_str(hexstr: &str) -> Result<Self, hex::FromHexError> {
+    pub fn from_hash_facsimile_str(hexstr: &str) -> Result<Self, hex::FromHexError> {
```
rtsisyk: May be "from_hex"? from_hash_facsimile_st is cumbersome

dbmcclain: I want something to specify that you are indeed expecting to import a hash image. From hex is still too broad. How about from_hash_str() ?

dbmcclain: BTW... can't you rename functions on import? as...?

rtsisyk: may be parse_hex()?

PR #128, opened 2018-11-06
change name to from_hex()

dbmcclain:
patch up to use from_hex()

make PBC abide by synthetic_random for choosing keys

add unit test to verify that PRNG is not being re-init on each call to functions that use thread_rng()

Add keying module to crypto lib - this will handle conversions to/from keys and wordlists

Add wordlist to/from bignum support

### DIFF
--- a/blockchain/src/genesis.rs
+++ b/blockchain/src/genesis.rs
@@ -66,20 +66,17 @@ pub fn genesis_dev() -> (Block, Vec<MerklePath>) {
 
     // Fool-proof checks.
     static PREVIOUS_HEX: &str = "daeed6308874de11ec5ba896aff636aee60821b397f88164be3eae5cf6d276d8";
-    static SKEY_HEX: &str = "daeed6308874de11ec5ba896aff636aee60821b397f88164be3eae5cf6d276d8";
-    static PKEY_HEX: &str = "bd2f2d45a439eafb3523216a652344883b3930f634a8c6e72eda55ff1f8670f9c90f139d9f9a486c6def760d2ff4e74d1f468c848c9774e63cdd0d46917eefe401";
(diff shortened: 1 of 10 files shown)


blockchain/src/genesis.rs
```diff
@@ -66,20 +66,17 @@ pub fn genesis_dev() -> (Block, Vec<MerklePath>) {
 
     // Fool-proof checks.
     static PREVIOUS_HEX: &str = "daeed6308874de11ec5ba896aff636aee60821b397f88164be3eae5cf6d276d8";
-    static SKEY_HEX: &str = "daeed6308874de11ec5ba896aff636aee60821b397f88164be3eae5cf6d276d8";
-    static PKEY_HEX: &str = "bd2f2d45a439eafb3523216a652344883b3930f634a8c6e72eda55ff1f8670f9c90f139d9f9a486c6def760d2ff4e74d1f468c848c9774e63cdd0d46917eefe401";
-    static SIG_HEX: &str = "901fc4370d2b19da0c4663ab4394d3e3cb74fdca7fc0ed7c8a8f0db3d2bff7d200";
+    static SKEY_HEX: &str = "104305c39c7f664feb4ac92c10b898eb8854645454640025732c9de7902e790e";
+    static PKEY_HEX: &str = "b05872990364a0bd71087ce252732e1b4370beb78e9f94aa8415c412aafb4142689043c1832e0df0fff3c10eb845dbf80b0621fd373fe2d2f3402cf56574cf8c00";
```
rtsisyk: I see that this linking to HEX values makes sense.

dbmcclain: not sure what you are stating. But I probably agree. BTW... I had to change your hex strings because of keying changes to deterministic randomness in crypto lib. Just made sure your unit tests passed okay with new keying...

rtsisyk: I meant that these checks add additional level of protection against accidental changes.

PR #131, opened 2018-11-06
Refactor network code

### LINKED ISSUE #122: Refactor network code
- Try to get rid of closures
- Clean up stegos.rs
- Implement a skeleton for randhound
- Switch to single-thread execution

Needed for #33

rtsisyk:
- Get rid of closures in main() and use tokio's futures explicitly.
- Extract command line handling into a separate service.
- Remove implicit subscriptions to topics.
- Replace node.send_to_id() with node.publish().
- Implement a stateful service for RandHound++.

Closes #122

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2000,6 +2000,8 @@ dependencies = [
  "clap 2.32.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "dirs 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.5.13 (registry+https://github.com/rust-lang/crates.io-index)",
+ "failure 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


rtsisyk:
Notes for Reviewer:

I tried "future.select2()" magical stuff to combine timer and floodsub events. It works, but I found that the resulting implementation is more complicated and less readable. I use low-level code which is simple and stupid.  My goal was to implement explicit timer/message handling without tons of closures, zillion levels of generics for resulting data types, move ||, .map_err() and similar garbage from the Tokio's universe.

rtsisyk:
BTW, you don't have to wait X seconds for the state transition. on_message() is triggered immediately when a new message is received. Just perform the state transition here and it will be highly interactive. I added a timer as an example, because it is needed at least to implement timeout. It is even possible to implement  synchronous calls on the top of unicast()/on_message(). Erlang does it in the same way (see gen_server:call()).

PR #133, opened 2018-11-07
Adapt the new approch of creating encrypted payload.

### LINKED ISSUE #121: Implement EncryptPayload

rtsisyk:
- Use Curve1174 crypto for transactions.
- Implement EncryptedPayload for MonetaryBlocks.
- Remove the list of witnesses from block.
- Refactor block tests.

Closes #121


blockchain/src/block.rs
```diff
@@ -71,16 +72,12 @@ pub struct Block {
     pub inputs: Vec<Input>,
 
     /// The list of transaction outputs in a Merkle Tree.
-    // TODO: replace with Merkle tree
     pub outputs: Merkle<Box<Output>>,
+    // TODO: BLS Multi-signature.
+    // pub sig: BlsSignature,
```
rtsisyk: I temporary commented out this line because I don't know how to create this signature.

blockchain/src/block.rs
```diff
@@ -138,96 +131,30 @@ impl Block {
         // Finalize the block hash
         let hash = hasher.result();
 
+        // Sign header
+        // TODO: how this signature should be used?
+        let _sig = sign_hash(&hash, skey);
```
rtsisyk: @dbmcclain I need you help here. I have block's hash and a private key. I guess that I need to sign hash using this private key. Since private key is from Curve1174, I'll get SchnorrSig here. How to use created SchnorrSig? What should be done with it?

blockchain/src/input.rs
```diff
@@ -21,21 +21,22 @@
...
-use stegos_crypto::pbc::secure::Signature;
 
 /// Transaction Input.
 #[derive(Debug, Clone)]
 pub struct Input {
     /// Identifier of an unspent transaction output.
     pub source_id: Hash,
     /// Signature used to sign transactions.
-    pub signature: Signature,
+    // TODO: which kind of signature should be used here?
```
rtsisyk: @dbmcclain

I guess that Input should contain some kind of signature signed by sender. Is it correct? Shall I use SchnorrSig here? I double-checked YellowPaper but haven't found an answer. Vlad also told me that input should have a hash of spent UTXO and some signature. Could you please explain?

crypto/src/pbc/secure.rs
```diff
@@ -567,12 +567,13 @@ impl fmt::Display for Signature {
     }
 }
 
-impl Hashable for Signature {
-    fn hash(&self, state: &mut Hasher) {
-        "Sig".hash(state);
-        self.base_vector().hash(state);
-    }
-}
+// NOTE: BLS Multi-signature never contributes to the computation of any block header hash.
```
rtsisyk: I intentionally commented this block out in order to prevent misuse of BLSSignature.hash().

blockchain/src/block.rs
```diff
@@ -24,9 +24,10 @@
 use input::Input;
 use merkle::*;
 use output::Output;
```
dbmcclain: Well, first off, I had already translated your old version to use Curve1174 and the new EncryptedPayload. So not sure why you are re-doing it...

But secondly, Blocks will use PBC crypto and BLS Multi-Signatures, not Curve1174. Encrypted payloads live inside of UTXOS, and those use single-curve ECC with Curve1174.

so let me read further what you have done...

blockchain/src/block.rs
```diff
@@ -91,8 +88,7 @@ impl Block {
         previous: Hash,
         timestamp: u64,
         leader: PublicKey,
```
dbmcclain: quite correct - adjustment - the sum of all gamma found in UTXOS will be from field Fr.

blockchain/src/block.rs
```diff
@@ -138,96 +131,30 @@ impl Block {
         // Finalize the block hash
         let hash = hasher.result();
```
dbmcclain: not sure why you are signing the hash of the header here... This will actually be a BLS MultiSignature, formed as the sum of individual BLS signatures from the Cosi witness group.

blockchain/src/genesis.rs
```diff
@@ -26,57 +26,72 @@ use chrono::prelude::{TimeZone, Utc};
...
 use stegos_crypto::hash::Hash;
-use stegos_crypto::pbc::fast::Zr;
-use stegos_crypto::pbc::secure::*;
 
 /// Genesis block for tests and development purposes.
 pub fn genesis_dev() -> (Block, Vec<MerklePath>) {
     let version: u64 = 1;
     let amount: i64 = 1_000_000;
     let epoch: u64 = 1;
     let previous = Hash::digest(&"dev".to_string());
```
dbmcclain: correct again... the delta cloaking factor on the recipient public key will be from field Fr. That's part of Curve1174 single-curve ECC

blockchain/src/payload.rs
```diff
@@ -21,55 +21,32 @@
...
+    // Convert amount to BE vector.
+    use std::mem::transmute;
+    let amount_bytes: [u8; 8] = unsafe { transmute(amount.to_be()) };
 
-impl fmt::Display for EncryptedPayload {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        fmt::Debug::fmt(self, f)
-    }
-}
+    let gamma_bytes: [u8; 32] = gamma.bits().to_lev_u8();
```
dbmcclain: this is what I call Lev32. A little-endian representation of a 256-bit number, with raw type [u8;32]. There is an entire module in our stegos_crypto crate to support type Lev32.

blockchain/src/payload.rs
```diff
@@ -21,55 +21,32 @@
...
-    pub fn garbage() -> EncryptedPayload {
-        let (_, pkey, _) = make_deterministic_keys(b"A test seed");
-        let pkg = aes_encrypt(b"This is a test", &pkey).unwrap();
-        EncryptedPayload {
-            apkg: pkg.apkg,
-            ag: pkg.ag,
-            cmsg: pkg.ctxt,
-        }
-    }
+    // String together a gamma, delta, and Amount (i64) all in one long vector and encrypt it.
```
dbmcclain: BTW... it is important that the PKey used here is the uncloaked version of the recipient's public key, so that he can decrypt and find the cloaking factor, delta, that was used in the published pkey of the UTXO.

crypto/src/curve1174/cpt.rs
```diff
@@ -240,12 +266,40 @@ pub fn validate_sig(hmsg: &Hash, sig: &SchnorrSig, pkey: &PublicKey) -> Result<b
 
 use std::iter::repeat;
 
+#[derive(Clone)]
 pub struct EncryptedPayload {
     pub apkg: Pt,
     pub ag: Pt,
     pub ctxt: Vec<u8>,
 }
```
dbmcclain: sorry, I thought I had already provided these hashable and debug prints. but perhaps not.

crypto/src/pbc/secure.rs
dbmcclain: completely incorrect. Blocks contains BLS Multi-signatures, UTXOS do not. UTXOs are the only things that use single-curve crypto on Curve1174.

blockchain/src/block.rs
dbmcclain: that signature comes, not from a function call, but from a Cosi consensus run between the leader node, who proposes the block, and all the Cosi witnesses who validate the proposal and individually sign the hash of the block. The leader then adds together all the signatures, and records in a bitmap which witnesses signed. That combination of a BLS multi-signature and the key bitmap makes the final block signature.

blockchain/src/block.rs
dbmcclain: again.. this is not a simple function call for obtaining a block multi-signature. It is a Cosi consensus protocol dance between the leader node and all the witnesses

blockchain/src/input.rs
dbmcclain: Inputs don't carry signatures. They are simply UTXO ID, a simple hash value (256 bits). It is up to the recipient (claimant) of a UTXO to furnish a verifiable ownership signature for the UTXO. Cosi witnesses will examine the spending transaction and validate the ownership signature.

blockchain/src/payload.rs
dbmcclain: BTW... little-endian for Curve1174, big-endian for PBC crypto. The distinction is important.

blockchain/src/input.rs
dbmcclain: let me refine that answer a bit... In a transaction, an Input will be a UTXO ID accompanied by the spender's proof of ownership signature. But the Block will only record the UTXO ID being spent. The Cosi witnesses validate the ownership signature before agreeing to sign the proposed transaction. But that signature gets stripped from the accepted transaction and only the input UTXO ID is recorded in the Block.

blockchain/src/block.rs
rtsisyk: In order to sign a block using BLS I need a block instance itself. In order to create a block I need to store some value to BlsSignature field. I can't store NULL here.

According to yellow paper (page 6, upper right part),  "The initial contents of the BLS MultiSignature slots should be zero fill end, even after being filled in, never contribute to the computation of any block header hash.".

I plan to apply this approach here. Block::new() will create zero-filled BlsSignature and save it in the block header. Cosi will sign the block and update the signature. Is it correct?

blockchain/src/block.rs
rtsisyk: Of course, I understand that we can't create Cosi signature here. I was misled with this sign_hash() line. I think that I should remove sign_hash() from Block::new() and don't require leader's private key in the block's constructor. Let me rework it.

blockchain/src/input.rs
rtsisyk: Thanks for the explanation! Now it is clear for me. It seems that we don't need the list of Inputs inside MonetaryBlock at all. The Input is only needed for transactions.

crypto/src/pbc/secure.rs
rtsisyk: What is wrong here? I added this comment just to note that BLSSignature NEVER needs Hashable implementation. We never hash BLSSignature, rights?

blockchain/src/input.rs
dbmcclain: I think you do need to keep a list of TXIN UTOX ID's in the Block, so that they can be matched up with later spend-side UTXO's and both become trimmed from the blockchain, leaving behind only their hash contributions to the Merkel trees.

rtsisyk:
Force-pushed. Please review.

blockchain/src/block.rs
```diff
@@ -71,28 +68,21 @@ pub struct Block {
     pub inputs: Vec<Input>,
```
rtsisyk: It seems that we don't need to store Inputs in blocks at all.

blockchain/src/input.rs
rtsisyk: > The only reason to hold TXIN UTXO ID's in a block is if you defer blockchain trimming to a later time. 

Yeah, I've got it too. I will remove Vec<Inputs> from Block and pass this  together with the block in order to prune old UTXOs.

blockchain/src/payload.rs
dbmcclain: From the Twist thread that I just posted:
```
As far as transmission and storage of crypto items go, they are simply [u8] vectors of varying size. So little-endian / big-endian have no meaning there. They should be stored and transmitted across the network as [u8] vectors.

Little-endian, Big-endian, only have meaning when viewing a succession of [u8] elements as a whole thing.
```

crypto/src/pbc/secure.rs
dbmcclain: Sorry, I was remarking about the comment in the code. For Blocks we really do use PBC and BLS signatures. But not single signatures, BLS Multi-Signatures. I thought you were meaning that we should be using Curve1174. So, probably my misunderstanding of what you meant.

blockchain/src/payload.rs
rtsisyk: OK

blockchain/src/block.rs
rtsisyk: I will remove Vec<Input> from the by following commits.

dbmcclain:
We have printing defined on all the crypto structs. So perhaps you could just borrow redirect debug to use those?

PR #134, opened 2018-11-07
Switched to log4rs for logging

echupriyanov:
* logging is configured by `stegos-log4rs.toml` file in current directory
* path to logging configuration file can be changed in main config file in [general] section
* if configuration file is bad or missing, default logging setup if configured in `initialize_logger` function


rtsisyk:
will review today.

config/src/types.rs
```diff
@@ -36,13 +36,32 @@ use rand::{thread_rng, Rng};
 pub struct Config {
     /// Network configuration.
     pub network: ConfigNetwork,
+    /// General settings
```
rtsisyk: should be before Network

src/stegos.rs
```diff
@@ -75,14 +78,30 @@ fn load_configuration(args: &ArgMatches) -> Result<Config, Box<Error>> {
...
+            let config = LogConfig::builder()
+                .appender(Appender::builder().build("stdout", Box::new(stdout)))
+                .logger(Logger::builder().build("stegos_network", LevelFilter::Debug))
+                .build(Root::builder().appender("stdout").build(LevelFilter::Info))
+                .expect("console logger should never fail");
+            log4rs::init_config(config)?
+        }
+    };
+
+    Ok(handle)
```
rtsisyk: Don't return this handle if it is never need.

rtsisyk:
There is another problem which need to be solved: `cargo test` doesn't display `log` output.

PR #139, opened 2018-11-08
Implement KeyBlocks

### LINKED ISSUE #124: Implement KeyBlocks
Stegos blockchain is able to accommodate multiple kinds of blocks. The kind of the block is defined by the Block Type ID slot in the block header. In this paper we define two types of blocks: 0(KeyBlock) and 1(M onetaryBlock). Monetary blocks, the blocks that carry all cryptocurrency transactions data and coins, will be described in details in Section VII-B. We will be referring to them simply as blocks throughout this paper. 

The purpose of Key blocks is to deliver the administrative information to blockchain participants. Any node, searching through key blocks, can see who is current or past epoch leader, who are witnesses (validators) for the current or past epoch. Nodes wishing to submit a transaction should consult with the latest key block to find who is the current pooled transactions facilitator.

rtsisyk:
Closes #124


dbmcclain:
just did a cargo clean and total rebuild. now the errors are gone. so approved.

PR #140, opened 2018-11-09
Rework UTXO

rtsisyk:
- Implement public key cloaking
- Remove computed hash value from Output
- Minor refactoring

See  #41


rtsisyk:
@dbmcclain please validate key cloaking: 

- https://github.com/stegos/stegos/blob/8d63cabc8c6a06db284ba5c772253a021025a54e/crypto/src/curve1174/cpt.rs#L146-L150
- https://github.com/stegos/stegos/blob/8d63cabc8c6a06db284ba5c772253a021025a54e/blockchain/src/output.rs#L80-L97

dbmcclain:
I don't think an application specific function belongs in a general purpose provider library. Your key cloaking should not reside in crypto/curve1174/cpt.rs. Key cloaking is a specific way of dealing with information to be published in a UTXO. So the cloaking code belongs there.

dbmcclain:
Furthermore... something you should become aware of, but you can't be blamed for not knowing it yet...

Public keys, presented in Pt compressed format, have the possibility of being completely invalid keys. The point may have been mangled on transmission, or might be junk inserted by attackers. So when expanding a compressed point in Pt form to ECp full form, it is possible that the point is not even on the curve, or belongs to a special small subgroup favored by attackers. The conversion from Pt to ECp can fail. And so it is not reasonable to use expect() because that could actually happen to a running system and we should be prepared to deal with it without killing the running program.

dbmcclain:
but otherwise, it looks good. You did the correct thing, using deterministic randomness, and I like the way you mix in a little bit of confusion with the timestamp.

rtsisyk:
> I don't think an application specific function belongs in a general purpose provider library. Your key cloaking should not reside in crypto/curve1174/cpt.rs. Key cloaking is a specific way of dealing with information to be published in a UTXO. So the cloaking code belongs there.

I agree. For this reason I split this logic into two parts:

curve1174::cpt:
```
   pub fn cloak(&self, delta: Fr) -> Self {
        let pt = Pt::decompress(self.0).expect("PublicKey is valid");
        PublicKey(ECp::compress(pt + delta * (*super::G)))
    } 
```

Application: actual logic with synthetic_random.

If you wish I can move `pub fn cloak(&self, delta: Fr) -> Self` into application, but it will also leak all Pt/ECp abstractions into application.

> Public keys, presented in Pt compressed format, have the possibility of being completely invalid keys. 

OK, I will throw CurveError in this case.

dbmcclain:
I don't see a problem with broad knowledge of ECp structure. That represents a full (X,Y) ECC point. Pt merely represents the convenient compressed form of a point. But Pt cannot be used for arithmetic on the curve, only ECp can be used for that.

dbmcclain:
I guess the "problem" with knowledge of ECp is its internal structure. Nobody should mess with that, treating ECp as an immutable opaque type. But if I make mod private, then all the other modules in curve1174 will break. So we must use gentlemen's agreement to see ECp as opaque. (or maybe there is a better way in Rust?)

rtsisyk:
> I guess the "problem" with knowledge of ECp is its internal structure. Nobody should mess with that, treating ECp as an immutable opaque type. But if I make mod private, then all the other modules in curve1174 will break. So we must use gentlemen's agreement to see ECp as opaque. (or maybe there is a better way in Rust?)

Try `pub(crate)` syntax.

According to a note about invalid PublicKeys. I think it would be better to validate all public keys on input, i.e. inside PublicKey::from_str(), PublicKey::from_vector() and so on. In this case are PublicKey instances will be valid.

> I don't see a problem with broad knowledge of ECp structure. That represents a full (X,Y) ECC point. Pt merely represents the convenient compressed form of a point. But Pt cannot be used for arithmetic on the curve, only ECp can be used for that.

I can move PublicKey::clock() into application if you wish.

Issue #126, opened 2018-11-05
One stream per topic


rtsisyk:
Allow to subscribe to individual topics.

PR #142, opened 2018-11-10
Switch to new network API

rtsisyk:
Follow up #126

### DIFF
--- a/randhound/src/lib.rs
+++ b/randhound/src/lib.rs
@@ -30,13 +30,16 @@ extern crate stegos_crypto;
 extern crate stegos_network;
 extern crate tokio;
 extern crate tokio_timer;
+#[macro_use]
+extern crate log;
 
 use failure::Error;
+use futures::sync::mpsc::UnboundedReceiver;
 use futures::{Async, Future, Poll, Stream};
 use std::thread;
 use std::thread::ThreadId;
 use std::time::Duration;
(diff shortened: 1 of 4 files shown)


randhound/src/lib.rs
```diff
@@ -60,25 +65,34 @@ pub struct RandHoundService {
...
+    fn on_unicast(&mut self, msg: Vec<u8>) {
+        info!(
+            "received unicast message: {}*",
+            String::from_utf8_lossy(msg.as_slice())
+        );
+    }
+
+    /// Called on a new broadcast message.
+    fn on_broadcast(&mut self, msg: Vec<u8>) {
+        info!(
```
echupriyanov: I'd suggest `debug!` here.

randhound/src/lib.rs
```diff
@@ -60,25 +65,34 @@ pub struct RandHoundService {
...
+        self.broker.publish(&TOPIC.to_string(), data)
     }
 
-    /// Called on a new message.
-    fn on_message(&mut self, msg: Vec<u8>) {
-        println!(
-            "randhound: *message: {}*",
+    /// Called on a new unitcast message.
+    fn on_unicast(&mut self, msg: Vec<u8>) {
+        info!(
```
echupriyanov: Let's replace it with `debug!()`

PR #159, opened 2018-11-14
Implement basic key chain

rtsisyk:
- Store public and secret key on disk using PEM format.
- Add a function to create CoSi keys from Wallet keys.
- Generate initial keys for testing.


keychain/src/lib.rs
```diff
@@ -21,10 +21,142 @@
...
+    pub cosi_skey: secure::SecretKey,
+    /// CoSi Public Key.
+    pub cosi_pkey: secure::PublicKey,
+    /// CoSi Signature.
+    pub cosi_sig: secure::Signature,
+}
+
+#[derive(Debug, Fail)]
+pub enum KeyChainError {
+    #[fail(display = "Failed to parse key: {}.", 0)]
```
echupriyanov: use _0 instead of 0

rtsisyk: Fixed and force-pushed.

rtsisyk:
Squashed and force-pushed.

PR #163, opened 2018-11-15
Fix failure of group leader to exit after sending randomness to Beaco…

dbmcclain:
I noticed that the group leader code keeps running in its loop, scanning over a batch, even after it has determined that it is time to finish and send the results to the Beacon node.

Now fixed. Also minor refactoring, and comment grammar corrections.


dbmcclain:
just rebased again off the most recent dev branch

PR #185, opened 2018-11-17
Clean up and update dependencies

rtsisyk:
- Switch to use `package = major.minor` syntax in Cargo.toml -
  full versions are already locked by Cargo.lock and it makes absolute
  no sense to lock patch-level manually.
- Move examples to appropriate workspace' folders.
- Remove unused dependencies.
- Update to the latest versions when possible.
- Run `cargo update`.


rtsisyk:
I'll rebase it after merging #186

PR #194, opened 2018-11-20
Reworked Randhound global state

echupriyanov:
* Load node's public/private keys from KeyChain.
* Load participants list and leader's public key from config files.


Cargo.lock
```diff
@@ -2148,6 +2148,11 @@ name = "stable_deref_trait"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[[package]]
+name = "state"
```
rtsisyk: This create is not used by this patch.

PR #196, opened 2018-11-21
Forbid invalid Pt and PublicKeys

### LINKED ISSUE #157: Forbid invalid Public/Secret keys in stegos_crypto
Add checks to all available PublicKey/SecretKeys constructors (from_str()/from_bytes() in order to validate that key is actually can be represented on the curve. Goal is to forbid invalid PublicKey/Secret keys object completely.

rtsisyk:
- Refactor Pt and PublicKeys constructors to forbin invalid points
- Unify into()/from() API for Pt, ECp, PublicKey, Fq, Fr, SecretKey
- Move PublicKey::cloak() into blockchain::Output as requested by David

Closes #157

### DIFF
--- a/crypto/src/curve1174/ecpt.rs
+++ b/crypto/src/curve1174/ecpt.rs
@@ -23,6 +23,7 @@
 //
 
 use super::*;
+use crate::CryptoError;
 
 // -----------------------------------------------------------------------------------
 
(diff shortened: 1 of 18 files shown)


crypto/src/curve1174/ecpt.rs
```diff
@@ -93,8 +93,8 @@ impl ECp {
         pt
     }
 
-    pub fn try_from(pt: Pt) -> Result<Self, CurveError> {
-        let mut x = pt.bits();
+    /// Try to convert from a compressed point.
```
echupriyanov: Looks like convert from raw bytes

crypto/src/curve1174/ecpt.rs
```diff
@@ -150,12 +157,22 @@ impl ECp {
         Self::from_random_bits(&random::<[u8; 32]>())
     }
 
-    pub fn compress(pt: Self) -> Pt {
-        Pt::from(pt)
+    /// Convert into compressed point.
```
echupriyanov: Looks like convert into raw bytes

crypto/src/curve1174/fields.rs
```diff
@@ -181,28 +181,43 @@ macro_rules! field_impl {
...
+            /// Convert from raw bytes.
             pub fn from_lev_u8(bytes: [u8; 32]) -> Self {
                 $name::Unscaled(U256::from_lev_u8(bytes))
             }
 
-            pub fn nbr_str(&self) -> String {
-                let tmp = (*self).unscaled_bits();
+            /// Convert from raw bytes.
+            #[inline]
+            pub fn from_bytes(bytes_slice: &[u8]) -> Self {
```
echupriyanov: why not `try_from_bytes`?

PR #198, opened 2018-11-22
Unify into_hex()/try_from_hex() API in crypto

rtsisyk:
Follow up #157

### DIFF
--- a/crypto/src/pbc/secure.rs
+++ b/crypto/src/pbc/secure.rs
@@ -105,16 +105,18 @@ impl Zr {
         Self::acceptable_random_rehash(x)
     }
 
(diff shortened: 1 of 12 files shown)


crypto/src/pbc/secure.rs
```diff
@@ -290,7 +301,7 @@ impl G2 {
 
 impl fmt::Display for G2 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.to_str())
+        write!(f, "{}", self.into_hex())
```
echupriyanov: add value type *fast::G2()"

PR #202, opened 2018-11-26
Working single run of Randhound

### LINKED ISSUE #200: Get single successful Randhound round
Subtask of #160

echupriyanov:
* Moved GlobalState from global var to internal state of Randhound Service
* Fixed a number of dead locks
* Added OOB message processing during stage transition

Closes #200 

Nots:
* it will crash after Final Lottery Ticket is obtained
* Needs to be carefully reworked to play nicely with Tokio runtime
* Be more resilient for network issues


rtsisyk:
OK

PR #212, opened 2018-12-04
Serialize genesis blocks

### LINKED ISSUE #125: UTXO hash is non-deterministic
UTXO hash contains Bulletproofs hash which are non-deterministic:

bulletproofs::make_range_proof(1_000_000) -> BP(vcmt: Pt(014119eb6f4e98396718607af6ebf529b7fb05a0911099ff4f3f70e369beeed7), ...)
bulletproofs::make_range_proof(1_000_000) -> BP(vcmt: Pt(042fc811d25f1526561c064cecfee41bda0f5efbf05a5a73844838c0b0e1acf0), ...)

How to solve this problem?

rtsisyk:
Add a tool to generate initial keys and blocks.

Closes #125


create-testing-keys.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+
+set -e
+NUM_KEYS=3
```
echupriyanov: NUM_KEYS=${1:-3}

This way it will be first argument of the script, or 3 if none is provided.

blockchain/src/genesis.rs
```diff
@@ -21,63 +21,30 @@
...
+        let previous = Hash::digest(&"genesis".to_string());
         let base = BaseBlockHeader::new(version, previous, epoch, timestamp);
 
-        let witnesses = keys
+        let witnesses = keychains
             .iter()
-            .map(|p| p.cosi_pkey.clone())
+            .map(|k| k.cosi_pkey.clone())
             .collect::<Vec<cosi_keys::PublicKey>>();
         let leader = witnesses[0].clone();
```
echupriyanov: Let the leader be the **last** node in the generated list

create-testing-keys.sh
rtsisyk: Fixed

blockchain/src/genesis.rs
rtsisyk: Discussed and decided don't change leader to the last node.

Issue #45, opened 2018-10-09
Limit the maximum size of block 


rtsisyk:
- [ ] Limit the maximum size of block (2MB)

rtsisyk:
A new logic: a block should be created when 2MB of transactions are collected in mempool OR when propose_timer expired.

PR #222, opened 2018-12-06
Limit the maximum size of block

rtsisyk:
- Add approximate calculation of block size.
- Trigger block creation when mempool has enough transactions.

See #45

### DIFF
--- a/blockchain/src/block.rs
+++ b/blockchain/src/block.rs
@@ -186,6 +186,13 @@ impl KeyBlock {
         // Create the block
         KeyBlock { header }
     }
+
+    /// Returns approximate the size of a block in bytes.
+    pub fn size_of(&self) -> usize {
+        let mut r = std::mem::size_of::<KeyBlock>();
+        r += self.header.witnesses.len() * std::mem::size_of::<SecurePublicKey>();
+        r
+    }
 }
 
 impl Hashable for KeyBlock {
@@ -298,6 +305,16 @@ impl MonetaryBlock {
         }
 
         Ok(())
+    }
+
+    /// Returns approximate the size of a block in bytes.
+    pub fn size_of(&self) -> usize {
+        let mut r = std::mem::size_of::<MonetaryBlock>();
+        r += self.body.inputs.len() * std::mem::size_of::<Hash>();
+        for (output, _) in self.body.outputs.leafs() {
+            r += output.size_of();
+        }
+        r
     }
 }
 
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -139,6 +139,8 @@ const BLOCK_TOPIC: &'static str = "block";
 const MONETARY_FEE: i64 = 1;
 const DATA_UNIT: u64 = 1024;
 const DATA_UNIT_FEE: i64 = 1;
+// Maximum block size
+const MAX_BLOCK_SIZE: usize = 2 * 1024 * 1024; // 2MB
 
 #[derive(Clone, Debug)]
 enum NodeMessage {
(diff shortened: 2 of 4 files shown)


rtsisyk:
Re-based on the latest dev.

node/src/lib.rs
```diff
@@ -839,15 +850,45 @@ impl NodeService {
...
         let mut fee = 0i64;
-
         let mut inputs = Vec::<Output>::new();
         let mut inputs_hashes = Vec::<Hash>::new();
         let mut outputs = Vec::<Output>::new();
-        for tx in self.mempool.drain(..) {
+        //
+        // Process tx_count first transactions in FIFO order
+        // Note that std::vec::Vec<> is used as FIFO queue here intentionally
+        // because one memmove() (i.e. .drain()) should be faster and easier
```
vldm: Probably this part "because one memmove() (i.e. .drain()) should be faster and easier" looks like premature optimizations.

Remember that later this code should work with some kind of persistent databases.

node/src/lib.rs
```diff
@@ -839,15 +850,45 @@ impl NodeService {
...
+        let mut spent_size_of = 0;
+        let mut tx_count: usize = 0;
+        for tx in self.mempool.iter() {
+            if std::mem::size_of::<Block>() + spent_size_of + tx.size_of() >= MAX_BLOCK_SIZE {
+                break; // stop before reaching the limit.
+            }
+            spent_size_of += tx.size_of();
+            tx_count += 1;
+        }
+        assert!(tx_count > 0); // block limit is adequate.
```
vldm: Should mempool always contain atleast one transaction?

blockchain/src/block.rs
```diff
@@ -186,6 +186,13 @@ impl KeyBlock {
         // Create the block
         KeyBlock { header }
     }
+
+    /// Returns approximate the size of a block in bytes.
+    pub fn size_of(&self) -> usize {
+        let mut r = std::mem::size_of::<KeyBlock>();
```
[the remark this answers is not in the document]
rtsisyk: I agree that block size is calculated in hacky way here. Of course it varies depending on Rust version  and architecture. But are there any better options? 

These calculations are needed to implement some adequate limit of the block size. We obviously don't want very large blocks in our blockchain.  But it is not problem ot create, say, 2200KB block when having 2048KB limit. This 10% error is just not important.

Actually, there is no proper definition what is the block size. In-memory size in Rust data structures? Compact in-memory size? Protobuf-packed size?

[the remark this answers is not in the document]
rtsisyk: In other words, I can subtract sizeof(Vec), but does it make sense?

node/src/lib.rs
rtsisyk: I just wanted to say that I'm using Vec here as FIFO queue. Probably it would be better to use list-like structure with boxed transactions to avoid memmove() on .drain(), but I don't think that it actually makes sense in terms of performance.

node/src/lib.rs
rtsisyk: Why? Where to get this one transaction on the start?

rtsisyk:
> The new methods name size_of looks the same as std::mem::size_of and can confuse some of new developers. Can we rename this methods to something like size_in_bytes? We can move this into some trait, for example Storable.

I agree. Need less general name than Storable. Please suggest.

blockchain/src/block.rs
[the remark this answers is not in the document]
vldm: This sounds reasonable.

node/src/lib.rs
vldm: Then i didn't uderstand this asserts `assert!(tx_count > 0)`

rtsisyk:
Plan to rebase on #248 and add tests.

rtsisyk:
> Then i didn't uderstand this asserts assert!(tx_count > 0)

This asserts checks that we picked at least one transaction. I realized that we need to limit the size of transaction too...

rtsisyk:
This patch has died. I'll rework it and open a new pull request.

PR #220, opened 2018-12-06
Implement data UTXO

rtsisyk:
- Split Output into MonetaryOutput and DataOutput
- Allow empty merkle tree and blocks without outputs
- Add "msg" command to CLI

Other changes:

- Fix error handling in CLI
- Print epoch changing notifications in CLI

Closes #165

PR #225, opened 2018-12-07
Implement fee calculation

### LINKED ISSUE #168: Implement monetary fee calculation
Currently fee is not implemented at all.

rtsisyk:
Add fee both for monetary and data transactions.

Closes #168


rtsisyk:
Based on #220

Issue #167, opened 2018-11-15
Create TLS keys from wallet keys


rtsisyk:
In order to simply the things for end user I propose to generate TLS keys used by libp2p by hashing wallet's secret key.

PR #233, opened 2018-12-10
Generate network keys on node start.

vldm:
Closes: https://github.com/stegos/stegos/issues/167

This pull request remove network keys from config, it is full compatible with old logic (as far as PeerID is used only for logs, as far as i understand), node will generates new keypair for libp2p on every run.

Additiionally, this pull requests remove old testdata files, and changes test config.

## Drawback.

If we want to introduce some linkage between  validator and network part, than we should modify schema, and share network key, along with validator key, before network startup.


network/src/node/mod.rs
```diff
@@ -155,18 +152,20 @@ impl Network {
         bind_ip.append(Protocol::Tcp(config.bind_port));
 
         let listen_addr = bind_ip;
-        let private_key = key_from_file(&config.private_key)?;
-        let public_key = key_from_file(&config.public_key)?;
+
+        // Generate new key on startup, and init peer_id.
+        let keypair = SecioKeyPair::secp256k1_generated()
+                            .expect("Couldn't generate keypair for network.");
```
rtsisyk: Please propagate all errors to main() instead of wrapping them.

vldm: The one reason for putting unwraps there, is that this code run only at initialization, and only one possible way to handle error on this code, is a panic about error. (If you prefer to keep panics in one place - this is ok.)

crypto/src/pbc/secure.rs
```diff
@@ -421,6 +421,12 @@ impl SecretKey {
     }
 }
 
+impl Into<[u8; 32]> for SecretKey {
```
rtsisyk: I'd prefer to use into_bytes()/from_bytes() convention like in all other keys (see PublicKey in this file and in curve1174). We wanted to switch to use Into/From trait, but missing TryFrom in stable stopped us.

keychain/Cargo.toml
```diff
@@ -10,5 +10,10 @@ failure = "0.1"
 base64 = "0.9"
 lazy_static = "1.1"
 regex = "1.0"
+
+# TODO: use the paritytech repo after https://github.com/paritytech/rust-secp256k1/pull/14
+eth-secp256k1 = { git = "https://github.com/tomaka/rust-secp256k1", branch = "pub-rand" }
```
rtsisyk: Just use the same crate as used by libp2p.

keychain/src/lib.rs
```diff
@@ -21,6 +21,9 @@
 
 #![deny(warnings)]
 
+#[macro_use]
+extern crate failure;
```
rtsisyk: Just `use failure::format_err;` or something like that, it's Rust 2018 ;)

keychain/Cargo.toml
```diff
@@ -10,5 +10,10 @@ failure = "0.1"
...
+rand = "0.4"
```
rtsisyk: rand = "0.6" is available and used everywhere in this project

keychain/src/lib.rs
```diff
@@ -167,4 +172,26 @@ impl KeyChain {
 
         keychain
     }
+
+    /// Generate new secp256k1 keypair using KeyChain as seed.
+    pub fn generate_secp256k1_keypair(&self)
+        -> Result<(secp256k1::key::SecretKey, secp256k1::key::PublicKey), Error>
+    {
+        // seed generator, with validator key.
+        let seed: [u8; 32] = self.cosi_skey.into();
```
rtsisyk: I think it would be better to chain to the wallet_skey instead of cosi key. cosi_key is already derived from wallet_skey.

keychain/Cargo.toml
vldm: comments and version was copy-pasted from libp2p, i can remove the comment.

rtsisyk: OK, no problem

keychain/Cargo.toml
vldm: I use public interface from eth-secp256k1 that is depend on 0.4 version.

rtsisyk: OK

Issue #30, opened 2018-10-09
Stakes mocked up via configuration


rtsisyk:

rtsisyk:
Just hard-code them until #50 and #52 are implemented.

Issue #50, opened 2018-10-09
Staking and Unstaking Transactions


rtsisyk:

rtsisyk:
Blocked by #272

PR #249, opened 2018-12-12
Mock validator stake.

vldm:
This probably all we need to start implementing leader election.

I'd like to rename variables "witnesses" to more common "validators". 

Closes: https://github.com/stegos/stegos/issues/30

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -138,7 +138,7 @@ impl Node {
 pub struct EpochNotification {
     pub epoch: u64,
     pub leader: SecurePublicKey,
-    pub witnesses: Vec<SecurePublicKey>,
+    pub witnesses: Vec<ValidatorState>,
 }
 
 /// Send when message is received.
@@ -210,6 +210,36 @@ pub enum NodeError {
     TooLowFee(i64, i64),
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub struct ValidatorState {
+    public_key: SecurePublicKey,
+    stake: u64,
+}
+
+impl ValidatorState {
+    pub fn new(public_key: SecurePublicKey, stake: u64) -> ValidatorState {
+        ValidatorState { public_key, stake }
+    }
+    pub fn without_stake(public_key: SecurePublicKey) -> ValidatorState {
+        ValidatorState {
+            public_key,
+            stake: 0,
+        }
+    }
+
+    pub fn public_key(&self) -> &SecurePublicKey {
+        &self.public_key
+    }
+
+    pub fn stake(&self) -> u64 {
+        self.stake
+    }
+
+    pub fn set_stake(&mut self, new_stake: u64) {
+        self.stake = new_stake;
+    }
+}
+
 struct NodeService {
     /// Blockchain.
     chain: Blockchain,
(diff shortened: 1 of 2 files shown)


node/src/lib.rs
```diff
@@ -167,6 +167,32 @@ pub enum NodeError {
...
+pub struct ValidatorState {
+    public_key: SecurePublicKey,
+    stake: u64,
+}
+
+impl ValidatorState {
+    pub fn new(public_key: SecurePublicKey, stake: u64) -> ValidatorState {
+        ValidatorState { public_key, stake }
+    }
+    pub fn without_stake(public_key: SecurePublicKey) -> ValidatorState {
```
rtsisyk: What is about stakes without validators? After #50 is implemented stakes can be changed every time MonetaryBlock is registered. The list of validators is only changed when a new KeyBlock is registered (=epoch changes).

node/src/lib.rs
```diff
@@ -221,7 +247,7 @@ impl NodeService {
         let unspent = HashMap::new();
         let epoch: u64 = 1;
         let leader: SecurePublicKey = G2::generator().into(); // some fake key
-        let witnesses = Vec::<SecurePublicKey>::new();
+        let witnesses = Vec::<_>::new();
```
rtsisyk: What is the difference between Vec::new() and Vec::<_>::new()? Just curious.

node/src/lib.rs
vldm: `ValidatorState` is a structure that responsible to track only current validators list.

Should i introduce aditional map that will track all known nodes stakes? (I throught it would be better to keep this on https://github.com/stegos/stegos/issues/50 )

node/src/lib.rs
vldm: There no difference between this twos, thank you.

There are other syntaxis that i suddenly mixed up:
`let data: Vec<_> = some_abstract_iter.collect();`

node/src/lib.rs
rtsisyk: Yes, I think it would be better to use a separate map for tracking stakes. It will be used during working on #50.

node/src/lib.rs
```diff
@@ -138,7 +138,7 @@ impl Node {
 pub struct EpochNotification {
     pub epoch: u64,
     pub leader: SecurePublicKey,
-    pub witnesses: Vec<SecurePublicKey>,
+    pub witnesses: Vec<ValidatorState>,
```
rtsisyk: Should be Vec<SecuretPublicKey> to avoid data redundancy. Create a temporary Map<SecurePublicKey, stake> for elections if you need.

node/src/lib.rs
```diff
@@ -210,6 +210,36 @@ pub enum NodeError {
     TooLowFee(i64, i64),
 }
 
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub struct ValidatorState {
```
rtsisyk: Unneeded, see above.

node/src/lib.rs
```diff
@@ -224,8 +254,10 @@ struct NodeService {
     epoch: u64,
     /// Current epoch leader.
     leader: SecurePublicKey,
-    /// The list of witnesses public keys.
-    witnesses: Vec<SecurePublicKey>,
+    /// The list of current validators states.
+    witnesses: Vec<ValidatorState>,
+    /// Map of all actual nodes stakes.
+    stakes: HashMap<SecurePublicKey, u64>,
```
rtsisyk: ValidatorState can be removed because HashMap<SecurePublicKey, u64> solves this problems.

node/src/lib.rs
```diff
@@ -321,6 +355,12 @@ impl NodeService {
                 Block::KeyBlock(key_block) => {
                     info!("Genesis key block: hash={}", Hash::digest(&key_block));
                     let key_block2 = key_block.clone();
+
+                    // TODO: remove stakes initialisation.
+                    for node in &key_block.header.witnesses {
+                        self.stakes.insert(*node, 10);
```
rtsisyk: I believe that different stakes are required for proper testing.

PR #252, opened 2018-12-12
Validate input and output range hashes of block

rtsisyk:
- Use Merkle tree for block.header.outputs_range_hash calculation
- Use chained hash for block.header.inputs_range_hash calculation
- Add checks to block.validate()
- Add tests

Needed for #47

### DIFF
--- a/blockchain/src/block.rs
+++ b/blockchain/src/block.rs
@@ -221,19 +221,12 @@ impl MonetaryBlock {
         let inputs = inputs.iter().map(|o| o.clone()).collect::<Vec<Hash>>();
 
         // Create outputs tree
-        let mut hasher = Hasher::new();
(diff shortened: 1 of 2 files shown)


dbmcclain:
Pedersen commitments balancing is one aspect. Where are amounts actually validated? (Bulletproofs)

rtsisyk:
Bulletproofs are only validated in transactions. Let me add this check too.

rtsisyk:
Sorry, I was wrong. BulletProofs are already validated by Block::validate():
See https://github.com/stegos/stegos/blob/afecb6a59429e5f2a00184ce5b400210ad14f5c5/blockchain/src/block.rs#L300-L302

rtsisyk:
I added comments, rebased and pushed this pull request as d3560780410cfe85943202e66e8d81e0f9f29e38.

PR #253, opened 2018-12-13
Add operators Add for infix sums of G1+G1, G2+G2, PublicKey + PublicK…

dbmcclain:
…ey, Signature+Signature, to support creation of BLS Multi-Signatures

You need these operations to form BLS Multisignatures. A multi-signature is the sum of two (or more) Signatures, and the corresponding validation key is the sum of two (or more) PublicKeys which correspond to the individual signatures. Both of these in pbc::secure along curves FR256

### DIFF
--- a/crypto/src/pbc/secure.rs
+++ b/crypto/src/pbc/secure.rs
@@ -38,7 +38,7 @@ use rand::thread_rng;
 use rand::Rng;
 use std::cmp::Ordering;
 use std::hash as stdhash;
-use std::ops::Neg;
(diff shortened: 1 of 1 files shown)


dbmcclain:
I will be dumbfounded if this does not work. Randhound has already proven that the same exact code works in the context of curves AR160. This is just a facsimile of the same code, now in the context of the more secure curves FR256

rtsisyk:
OK, let's wait until CI finished.

PR #255, opened 2018-12-13
Integration of RandHound and Node services

echupriyanov:
* implemented RandHound API to be orchestrated ny Node
* added Randomness to the Node internal state
* start next RandHound round, when Randomness is obtained


node/src/lib.rs
```diff
@@ -528,6 +553,16 @@ impl NodeService {
         Ok(())
     }
 
+    fn handle_new_randomness(&mut self, h: Hash) -> Result<(), Error> {
+        self.randomness = h;
+        info!("New randomness obtained: {}", h);
+        // Start new round, if we are the leader
+        if self.keys.cosi_pkey == self.leader {
```
rtsisyk: just use self.is_leader()

randhound/src/lib.rs
```diff
@@ -199,6 +240,7 @@ impl<'a> RandHoundService {
         }
         self.state.set_next_epoch(epoch);
         // TODO: remove this in favor Node service orchestrating RandHound
+        // Remove this, when appropriate code is present in Node
```
rtsisyk: What's missing?

randhound/src/randhound.rs
```diff
@@ -651,9 +658,17 @@ impl GlobalState {
...
-            MsgType::FinalLotteryTicket { .. } => {
+            MsgType::FinalLotteryTicket { ref ticket } => {
+                info!("Final lottery ticket receiced: {}", ticket);
                 // There should me more to do here...
                 // (hold election, assign new roles, etc.)
+                for c in self.consumers.iter() {
+                    if let Err(e) = c.unbounded_send(Randomness {
+                        value: ticket.clone(),
+                    }) {
+                        error!("Failed to send generated Randomness to consumers: {}", e);
```
rtsisyk: I just silently removed consumer from the list:
```
                    self.on_message_received
                        .retain(move |tx| tx.unbounded_send(msg.clone()).is_ok());
```

randhound/src/randhound.rs
```diff
@@ -819,7 +834,17 @@ impl GlobalState {
             //
             let my_group = grps.first().unwrap();
             let my_leader = my_group.first().unwrap();
-            let session_id = Hash::digest_chain(&[&self.get_current_epoch(), &self.get_skey()]);
+            let mut hasher = Hasher::new();
+            let _secs_from_bigbang = SystemTime::now()
```
rtsisyk: What is it? Why system time is needed here?

randhound/src/lib.rs
```diff
@@ -137,12 +177,12 @@ impl<'a> RandHoundService {
 
         inputs.push(Box::new(broadcast_rx));
 
-        let epoch_rx = node
-            .subscribe_epoch()?
-            .map(|m| RandHoundEvent::Epoch(m))
-            .map_err(|_| RandHoundInputError::NoError);
+        // let epoch_rx = node
```
rtsisyk: Don't comment out, please remove instead.

randhound/src/randhound.rs
```diff
@@ -1606,6 +1631,11 @@ impl GlobalState {
                     let trand = fast::compute_pairing(&G1::generator(), &grand);
                     let ticket = Hash::digest(&trand);
                     info!("Calculated Final Lottery Ticket: {}", ticket);
+                    for c in self.consumers.iter() {
```
rtsisyk: I suggest to remove died subscribers from the list.

randhound/src/randhound.rs
```diff
@@ -651,9 +658,16 @@ impl GlobalState {
                 self.stash_subsubgroup_randomness(&from, &rands)
             }
             MsgType::GroupRandomness { ref rand } => self.stash_group_randomness(&from, &rand),
-            MsgType::FinalLotteryTicket { .. } => {
+            MsgType::FinalLotteryTicket { ref ticket } => {
+                info!("Final lottery ticket receiced: {}", ticket);
                 // There should me more to do here...
                 // (hold election, assign new roles, etc.)
```
vldm: Comment is probably outdated?
After node being notified, we should waiting for new epoch, or have i miss anything?

start-cluster.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-mode=${2:-release}
+mode=${2:-debug}
```
vldm: Is it important to change mode to `debug`?

node/src/lib.rs
```diff
@@ -220,6 +224,8 @@ struct NodeService {
     /// A monotonically increasing value that represents the heights of the blockchain,
     /// starting from genesis block (=0).
     epoch: u64,
+    /// Randomness, generated by RandHound
+    randomness: Hash,
```
vldm: Can we add some counter, to synchronize random number between nodes?

Something like 
`randomness: Vec<Hash>`, 
Or even `randomnes: (usize, Hash)` would be good.

start-cluster.sh
rtsisyk: Looks like unrelated change, but I just ignore that.

node/src/lib.rs
rtsisyk: I don't understand idea.

vldm: When we need to form some message based on random, we should to somehow synchronize that random number.

PR #256, opened 2018-12-13
Add infastructure for CoSi

rtsisyk:
- Add protobuf packets
- Use BTreeSet for the list of witnesses


blockchain/src/block.rs
```diff
@@ -167,15 +168,13 @@ impl KeyBlock {
...
-        let mut witnesses = witnesses.to_vec();
-
-        // Witnesses list must be sorted.
-        witnesses.sort();
+        // The set of witnesses must be non-empty.
+        assert!(!witnesses.is_empty());
 
         // Leader must present in witnesses array.
-        //assert_eq!(witnesses.binary_search(leader), Ok((_, _)));
+        assert!(witnesses.contains(&leader));
```
vldm: can we put comments to assert message ?
```suggestion
        assert!(witnesses.contains(&leader), "Leader must present in witnesses array.");
```

blockchain/src/block.rs
```diff
@@ -167,15 +168,13 @@ impl KeyBlock {
...
         leader: SecurePublicKey,
-        witnesses: &[SecurePublicKey],
+        witnesses: BTreeSet<SecurePublicKey>,
     ) -> Self {
-        let mut witnesses = witnesses.to_vec();
-
-        // Witnesses list must be sorted.
-        witnesses.sort();
+        // The set of witnesses must be non-empty.
+        assert!(!witnesses.is_empty());
```
vldm:
```suggestion
        assert!(!witnesses.is_empty(), "The set of witnesses must be non-empty.");
```

blockchain/src/genesis.rs
```diff
@@ -44,13 +45,13 @@ pub fn genesis(keychains: &[KeyChain], amount: i64) -> Vec<Block> {
         let previous = Hash::digest(&"genesis".to_string());
         let base = BaseBlockHeader::new(version, previous, epoch, timestamp);
 
-        let witnesses = keychains
+        let witnesses: BTreeSet<cosi_keys::PublicKey> = keychains
             .iter()
             .map(|p| p.cosi_pkey.clone())
-            .collect::<Vec<cosi_keys::PublicKey>>();
-        let leader = witnesses[0].clone();
+            .collect::<BTreeSet<cosi_keys::PublicKey>>();
```
vldm: You can omit this type parameter
```suggestion
            .collect();
```

blockchain/src/genesis.rs
```diff
@@ -44,13 +45,13 @@ pub fn genesis(keychains: &[KeyChain], amount: i64) -> Vec<Block> {
         let previous = Hash::digest(&"genesis".to_string());
         let base = BaseBlockHeader::new(version, previous, epoch, timestamp);
 
-        let witnesses = keychains
+        let witnesses: BTreeSet<cosi_keys::PublicKey> = keychains
```
vldm: and this line could be minified 
```suggestion
        let witnesses: BTreeSet<_> = keychains
```

blockchain/src/transaction.rs
```diff
@@ -159,6 +159,11 @@ impl Transaction {
     pub fn validate(&self, inputs: &[Output]) -> Result<(), Error> {
         assert_eq!(self.body.txins.len(), inputs.len());
 
+        // Check fee.
+        if self.body.fee < 0 {
```
vldm: Why we can't make `fee:u64`, and move this assert to type level?

blockchain/src/block.rs
rtsisyk: Thanks for suggestion! I didn't know that assert!() has second argument. I will update this commit.

blockchain/src/transaction.rs
rtsisyk: I think that David decided to use `i64` for money in order to prevent overflow during computations. Signed/unsigned comparison is well-known undefined behaviour, so I use i64 for money everywhere.

rtsisyk:
@echupriyanov @vldm I desperately need your review

PR #260, opened 2018-12-14
Replaced Delay futures with single DelayQueue for quorum deadline

echupriyanov:
Using Delay() future pollutes runtime with many unnecessary futures. Using DelayQueue is more straightforward and allows clearing delay, if all keys were received before deadline

### DIFF
--- a/randhound/src/lib.rs
+++ b/randhound/src/lib.rs
@@ -28,7 +28,7 @@ use crate::randhound::{EpochInfo, GlobalState};
 
 use failure::{Error, Fail};
 use futures::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
-use futures::{Async, Future, Poll, Stream};
+use futures::{self, Async, Future, Poll, Stream};
 use futures_stream_select_all_send::select_all;
 use log::*;
 use protobuf;
@@ -113,7 +113,6 @@ pub(crate) enum RandHoundEvent {
     Broadcast(Vec<u8>),
     Epoch(RandhoundEpoch),
     Heartbeat(HeartbeatUpdate),
-    CheckInitQuorum,
     Subscribe(UnboundedSender<Randomness>),
     NewRound,
 }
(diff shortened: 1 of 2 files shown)


randhound/src/lib.rs
```diff
@@ -304,12 +306,19 @@ impl Future for RandHoundService {
                     RandHoundEvent::Epoch(msg) => self.on_epoch(msg),
                     RandHoundEvent::Heartbeat(msg) => self.on_heartbeat(msg),
                     RandHoundEvent::Timer(_i) => self.on_timer(),
-                    RandHoundEvent::CheckInitQuorum => self.on_check_init_quorum(),
                     RandHoundEvent::NewRound => self.on_new_round(),
                     RandHoundEvent::Subscribe(tx) => self.state.subscribe(tx),
                 },
                 Ok(Async::Ready(None)) => unreachable!(), // never should happen
-                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Ok(Async::NotReady) => match self.state.poll_deadline() {
```
rtsisyk: Why deadlines are polled only when main loop doesn't have events? All timers and deadlines are usually more important when network events.

randhound/src/randhound.rs
```diff
@@ -306,13 +304,22 @@ pub(crate) fn init_state(
...
         msg_queue: VecDeque::new(),
-        runtime,
         service,
         consumers: vec![],
+        deadline: DelayQueue::new(),
     }
 }
 
 impl GlobalState {
+    pub fn poll_deadline(&mut self) -> Poll<Option<Hash>, Error> {
```
rtsisyk: I don't understand why this part of RandHoundService::poll() should be in GlobalService.   It's time to merge libs.rs with randhound.rs. As a temporary solution I can suggest to merge this code back to  RandHoundService::poll().

rtsisyk: Discussed and agreed to fix in a separate ticket.
https://github.com/stegos/stegos/issues/266

randhound/src/lib.rs
```diff
@@ -297,19 +299,35 @@ impl Future for RandHoundService {
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
+            let deadline_poll = self.state.poll_deadline();
```
vldm: If atleast one message was in `self.recv` when timeout expire, then we probably would ignore this deadline?

For example, imagine next flow:
1) `self.state.poll_deadline()` return's `Async::Ready(Some(...))` 
2) `self.recv.poll()` returns somethin other than `Async::NotReady`
After this next iteration of loop began and new call to `self.state.poll_deadline()` starts, old `deadline_poll` result will be forgetten.

Have i missing something?

echupriyanov: Result of deadline_poll() is processed in the match statement. On the next iteration we check if any more expirations happen

Issue #38, opened 2018-10-09
Validator Group Formation


rtsisyk:
- [ ] Maintain the list of alive staked nodes for randhound and validator group formation.
- [ ] Form validator group when epoch changes

PR #279, opened 2018-12-19
Create group of validators based on random.

vldm:
Closes: #38 

Choose a group of validators based on ranndhound number.


rtsisyk:
Let's merge it and improve by following pull requests.

PR #281, opened 2018-12-19
Initial consensus implementation for MonetaryBlocks

rtsisyk:
Initial consensus implementation for MonetaryBlocks

- Add implicit single-round CoSi state machine.
- Add MonetaryBlockProposal and BlockAcceptance messages.
- Validate transactions and block in MonetaryBlockRequest.
 - Collect BLS signatures and wait for BFT supermajority.
 - Commit proposed block after receiving BFT supermajority.
 - Check BLS Signatures.

Limitations:

- Consensus is implemented using the single round of CoSi.
- Mempool is completely discarded when a sealed block is received.

These limitations will be addressed by the following patches.

See #47


rtsisyk:
Added BLS signatures.

node/src/lib.rs
```diff
@@ -229,12 +273,26 @@ struct NodeService {
...
     leader: SecurePublicKey,
     /// The list of witnesses public keys.
     witnesses: BTreeSet<SecurePublicKey>,
     /// Map of all actual nodes stakes.
     stakes: HashMap<SecurePublicKey, i64>,
+    /// Hash of a block proposed by CoSi protocol.
+    proposed_block_hash: Option<Hash>,
+    /// A block proposed by CoSi protocol.
+    /// Sic: this field is only used by leader.
+    proposed_block: Option<Block>,
```
vldm: Probably `NodeService` now a kind of "god object" and we should refactor it.

For now we probably can split cosi state into enum, just to merge this option into one field.

```rust
struct BlockFormation {
  block_hash: Hash,
  block: Block,
  signatures: ...
  // some kind of statuses from witnesses can be append there
}

enum ConsensusState {
      Idle,
// we can seperate phase of transaction verify too.
      BlockFormation(BlockFormation),
      
}
```

PR #282, opened 2018-12-19
Add Signature and bitmap of witnesses to the block header

rtsisyk:
Just add to the block header. Validation is not implemented yet.

Needed for #42

### DIFF
--- a/blockchain/src/block.rs
+++ b/blockchain/src/block.rs
@@ -24,6 +24,7 @@
 use crate::error::*;
 use crate::merkle::*;
 use crate::output::*;
+use bitvector::BitVector;
 use failure::Error;
 use std::collections::BTreeSet;
 use std::collections::HashSet;
@@ -34,9 +35,12 @@ use stegos_crypto::curve1174::fields::Fr;
 use stegos_crypto::curve1174::G;
 use stegos_crypto::hash::{Hash, Hashable, Hasher};
 use stegos_crypto::pbc::secure::PublicKey as SecurePublicKey;
+use stegos_crypto::pbc::secure::Signature as SecureSignature;
+
+pub const WITNESSES_MAX: usize = 128;
 
 /// General Block Header.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct BaseBlockHeader {
     /// Version number.
     pub version: u64,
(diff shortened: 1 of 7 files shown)


crypto/src/pbc/secure.rs
```diff
@@ -200,10 +200,16 @@ impl Neg for Zr {
 pub struct G1([u8; G1_SIZE_FR256]);
 
 impl G1 {
-    pub fn new() -> G1 {
+    /// Create a new point which binary representation consists of all zeros.
+    pub fn null() -> G1 {
```
vldm: may be use `zero` name like in other crypto types ?
```suggestion
    pub fn zero() -> G1 {
```

crypto/src/pbc/secure.rs
```diff
@@ -673,6 +679,17 @@ impl PartialEq for PublicSubKey {
 pub struct Signature(G1);
 
 impl Signature {
+    /// Create a new point which binary representation consists of all zeros.
+    pub fn null() -> Signature {
```
vldm:
```suggestion
    pub fn zero() -> G1 {
```

blockchain/src/block.rs
```diff
@@ -34,9 +35,12 @@ use stegos_crypto::curve1174::fields::Fr;
 use stegos_crypto::curve1174::G;
 use stegos_crypto::hash::{Hash, Hashable, Hasher};
 use stegos_crypto::pbc::secure::PublicKey as SecurePublicKey;
+use stegos_crypto::pbc::secure::Signature as SecureSignature;
+
+pub const WITNESSES_MAX: usize = 128;
```
vldm: why we have this limit?

rtsisyk: Just some reasonable number. Can increase to 512 or more.

crypto/src/pbc/secure.rs
rtsisyk: I don't like both variants. zero() can be perceived as zero point on the curve... I think that this problem is caused by bad design of block header. Why we want to store "empty" signature and then change it?

blockchain/src/block.rs
vldm: Have technical limit, or it's just for testing?

rtsisyk:
I'm integrating this pull request to the main branch #281

PR #295, opened 2019-01-09
WIP: Implementing VRF tickets system.

vldm:


dbmcclain:
In function select_winner, you choose a random seed and then perform `random % sum_stakes`. That doesn't seem correct to me. Viewing random as a fractional value in the interval (0,1), why shouldn't this be `random * sum_stakes`?

vldm:
>In function select_winner, you choose a random seed and then perform random % sum_stakes. That doesn't seem correct to me. Viewing random as a fractional value in the interval (0,1), why shouldn't this be random * sum_stakes?

@dbmcclain  in this case i use integer arithmetic, and random value could be in range `[0, 2^64]` . So by doing `random % sum_stakes` we map a value to range `[0, SUM_STAKES]`. Do you thing there are problem in using integer arithmetic?

dbmcclain:
No, in retrospect I can’t challenge what you are doing… I have just never seen your approach before. Perhaps later I will find a proof that this is not valid, but for now I can’t do that.

dbmcclain:
still getting compile time syntax errors:
``
error[E0063]: missing field `queue` in initializer of `tickets::VRFTicketsSystem`
   --> node/src/tickets.rs:196:9
    |
196 |         VRFTicketsSystem {
    |         ^^^^^^^^^^^^^^^^ missing `queue`

error[E0308]: mismatched types
   --> node/src/tickets.rs:264:21
    |
264 |                     collecting.process_ticket(out_of_order_ticket)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^- help: try adding a semicolon: `;`
    |                     |
    |                     expected (), found enum `std::result::Result`
    |
    = note: expected type `()`
               found type `std::result::Result<(), tickets::VRFTicketsError>`

error: aborting due to 2 previous errors
``

vldm:
@dbmcclain Sorry i use this branch to build on remote server, so sometimes it could be even "Unbuildable", i will remove WIP prefix when everything could be builded and tested, for now i only want to see if there some "high level" issues with code.

dbmcclain:
Understood…

node/src/election.rs
```diff
@@ -87,23 +94,16 @@ pub fn choose_validators(
         let index = select_winner(stakers.iter().map(|(_k, stake)| stake), rand).unwrap();
```
rtsisyk: This algorithm is extremely inefficient:

1. `select_winner()` iterates `stakes` array from the start every time.
2. `stakes.remove(index)` memmoves the rest array on deleting.

Please consider to redesign algorithm and/or use appropriate data structures.

node/src/lib.rs
```diff
@@ -271,10 +277,10 @@ struct NodeService {
...
     /// starting from genesis block (=0).
     epoch: u64,
-    /// A configurable value of a maximum group size.
-    max_group_size: usize,
-    /// A current election round.
-    election_round: u32,
+
+    /// A system that restart consensus in case of fault or partition.
+    /// And allow to change validators in case of epoch change.
+    vrf_system: VRFTicketsSystem,
```
rtsisyk: The name `VRFTicketSystem` is too long. Try to find a shorter name.

node/src/protos/mod.rs
```diff
@@ -868,6 +870,46 @@ impl FromProto<node::ConsensusMessage> for ConsensusMessage {
...
+        Ok(VRFTicket {
+            random,
+            author,
+            sig,
+        })
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
rtsisyk: Please add a test for enconding/decoding. Protobuf boilerplate is error-prone and you can easily forget to serialize some fields here.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+use std::{
+    collections::HashMap,
+    mem,
+    time::{Duration, Instant},
+};
+
+#[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub struct VRFTicket {
+    pub random: VRF,
+    pub author: SecurePublicKey,
```
rtsisyk: ConsensusMessage and SealedBlockMessage are very similar to `VRFTicket` by structure, but they use `pkey` name instead of `author`. I decided to use name `pkey` instead of `sender`/`author` because the code operates both with `pkey` and `skey` at the same time in multiple place.  Shall we unify the structure?

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    fn hash(random: &VRF, author: &SecurePublicKey) -> Hash {
+        let mut hasher = Hasher::new();
+        random.hash(&mut hasher);
+        author.hash(&mut hasher);
+        hasher.result()
+    }
+
+    fn validate(&self, seed: Hash) -> bool {
```
rtsisyk: For similar cases I returned `Result<(), Error>` instead of bool. See Block::validate(), Transaction::validate() and other.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+            tickets: HashMap::new(),
+            seed: last_block_hash,
+        }
+    }
+
+    /// Creates new tickets
+    /// Panics if we failed to process new ticket.
+    fn produce_ticket(
+        &mut self,
+        public_key: SecurePublicKey,
```
rtsisyk: `pkey`/`skey` convention is widely used in the code base.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+            if let Some(old_random) = self.tickets.insert(ticket.author, random) {
+                if old_random != random {
+                    return Err(VRFTicketsError::MultipleTickets(ticket.author));
+                }
+            }
+            Ok(())
+        }
+    }
+
+    fn lowest(self) -> Result<VRF, VRFTicketsError> {
```
rtsisyk: Just use an ordered data structure like BTreeSet instead of HashSet.

BTW, why this method is consuming `self` and have `.cloned()` at the same time?

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+                } else {
+                    item
+                };
+                Some(value)
+            })
+            .ok_or(VRFTicketsError::NoTickets)
+    }
+}
+
+enum SystemState {
```
rtsisyk: May be VRFState or something like that?

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+    CollectingTickets(Collecting, Instant),
+}
+
+impl Default for SystemState {
+    fn default() -> SystemState {
+        SystemState::Sleeping(Instant::now())
+    }
+}
+
+pub struct VRFTicketsSystem {
```
rtsisyk: Please add comments to all members of public structures.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+        match self.state {
+            SystemState::Sleeping(start) if time.duration_since(start) > RESTART_CONSENSUS => self
+                .on_view_change(last_block_hash)
+                .map(Feedback::BroadcastTicket),
+            SystemState::CollectingTickets(_, start)
+                if time.duration_since(start) > COLLECTING_TICKETS =>
+            {
+                self.on_collection_end(stakers).map(Feedback::ChangeGroup)
+            }
+            _ => Ok(Feedback::Nothing),
```
rtsisyk: There are no more states. This case can be removed.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn handle_epoch_end(
+        &mut self,
+        last_block_hash: Hash,
+    ) -> Result<VRFTicket, VRFTicketsError> {
+        self.on_view_change(last_block_hash)
+    }
+
+    /// On receiving valid block, trying to restart VRF system.
+    pub fn handle_sealed_block(&mut self) {
+        self.view_change = 0;
```
rtsisyk: I don't see a mechanism to re-create validator group after N SealedBlocks.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+        self.state = Default::default();
```
rtsisyk: BTW, I think that self.state = SystemState::Sleeping(Instant::now()) is more readbale than Default::default(). Please just ignore this comment if you don't think so.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    /// Receive ticket from other nodes.
+    pub fn hanle_process_ticket(&mut self, ticket: VRFTicket) -> Result<(), VRFTicketsError> {
+        debug!("Receiving new ticket from = {:?}.", ticket.author);
+        match &mut self.state {
+            SystemState::CollectingTickets(ref mut state, _) => {
+                state.process_ticket(ticket)?;
+            }
+            _ => {
+                self.queue.push(ticket);
```
rtsisyk: How outdated tickets are handled? Tickets from the next epochs should be queued, but tickets from previous epoch should be ignored. 

I'm not sure that tickets from epoch + 1 should raise an OutOfOrderTicket. This case is not error, just asynchronicity.

node/src/tickets.rs
```diff
@@ -0,0 +1,307 @@
...
+    InvalidTicket(VRFTicket),
+    #[fail(display = "No tickets was received in time.")]
+    NoTickets,
+    #[fail(display = "Trying to process tickets not in collection phase.")]
+    OutOfOrderTicketsProcessing,
+    #[fail(display = "Receiving multiple tickets from {:?}.", _0)]
+    MultipleTickets(SecurePublicKey),
+}
+
+struct Collecting {
```
rtsisyk: I didn't get it at the first attempt. May be it should be renamed to VRFCollectingState or something like that?

node/src/tickets.rs
vldm: 1) I need ordered Values, not keys, to optimise this code we could save only lowest hash, but then we should deal with flood.

2) Consuming `self` is just semantic of this method (i force moving to other state after `lowest` is called).
`cloned()` for `Copy` types should be free, but i can rewrite it to use `into_iter().map(|(_,v)| v)` instead of `values()`

node/src/lib.rs
vldm: Fell free to suggest x)

We can ommit VRF, and use `TicketsSystem` elsewhere.

node/src/election.rs
vldm: Probably this would be a premature optimisation.

I can bench different algorithms in separate task, and we will decide the most good for us.

Complexety of selecting group is O(N*m), so it should scale good enough for reasonable N.

node/src/tickets.rs
vldm: I like the unification, but i didn't prefer `pkey` because it didn't carry any semantic. We already know that this is `PublicKey`, from type.

node/src/tickets.rs
vldm: It's not a state of `Verifiable Random Function`, probably it would better to omit `VRF` prefix elsewhere?

crypto/src/pbc/secure.rs
```diff
@@ -1019,12 +1019,19 @@ pub fn ibe_decrypt(pack: &EncryptedPacket, skey: &SecretKey) -> Option<Vec<u8>>
 
 // -------------------------------------------------------
 // VRF's provable, unforgeable, deterministic randomness
-
+#[derive(Eq, PartialEq, Debug, Clone, Copy)]
```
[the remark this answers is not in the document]
vldm: PartialEq is needed because i need to compare some VRFTicket with other one from user.

node/src/tickets.rs
vldm: i created a ticket for this https://github.com/stegos/stegos/issues/309

node/src/tickets.rs
vldm: This type is private, so i prefer to keep it without namespace. But State prefix looks good.

PR #336, opened 2019-01-16
Initial implementation of unicast communications

echupriyanov:
All nodes are subscribed to special topic `stegos-unicast`.
To send 'direct' message, node publishes message in that topic
accompanied with destination secure::PublikKey.

All nodes are picking messages from this topic, which have to: fields
equal to their cose_pkey

Added send <PUBLICKEY> <MSG> command to the REPL


src/console.rs
```diff
@@ -192,6 +196,13 @@ impl ConsoleService {
         println!("");
     }
 
+    fn help_send() {
```
rtsisyk: I think that we need to split dev and user commands in REPL. Anyway, let's accept it as it and rename later.

rtsisyk:
Please for the review from @vldm, he had some comments.

PR #337, opened 2019-01-16
Rework our testnet constants, to work faster.

vldm:


rtsisyk:
Probably all hard-coded constants of our blockchain (e.g. timeouts, fees, etc.) should be extracted into a separate file. But let's do it later.

PR #338, opened 2019-01-16
Implement Escrow UTXO

rtsisyk:


rtsisyk:
Re-based on the latest `dev`.

rtsisyk:
@dbmcclain please review

rtsisyk:
> I wonder if we should make the key cloaking take an Option(gamma) argument instead of insisting on gamma = 0? For those cases you would provide a None argument. Elsewhere when gamma comes from a Bulletproof you would supply Some(gamma). ??

Yes, I thought about that. I'll make a follow-up patch.

Issue #317, opened 2019-01-14
Validate block's timestamp


rtsisyk:
Block's timestamp must be in the interval +/- 10 minutes relative to local realtime clock.

Issue #273, opened 2018-12-17
Unit tests for consensus


vldm:
- [x] Implement object that will create a test Consensus enviroment:

Possible api:
- [x] Numbers of node to start
- [x] Receive msg from **external consensus Node**
- [x] Send message to **external consensus Node**
- [x] Get current state (stage, variables)
- [ ] Reset state machine

Definition of done: Test case should be implemented - 
1) Start consensus with 3 nodes; 
2) process through all stages; 
3) check that all nodes ends up in the same state.

rtsisyk:
s/Randhound/Consensus/g

rtsisyk:
New consensus logic can be fully tested via unit tests.

vldm:
This task was initially about testing infrastructure, so i am closing this as obsolete.

@rtsisyk If you have additional info please provide and reopen this task.

PR #340, opened 2019-01-16
Check timestamp of block in block proposal.

vldm:
Closes: https://github.com/stegos/stegos/issues/317

Add timestamp check to consensus propose.
We still need somehow handle timestamp of sealed block, but this should be a part of fork resolution.

And we need https://github.com/stegos/stegos/issues/273 to fully test this pull-request


node/src/lib.rs
```diff
@@ -181,6 +181,8 @@ const TX_WAIT_TIMEOUT: Duration = Duration::from_secs(30);
 const CONSENSUS_TIMER: Duration = Duration::from_secs(30);
 /// Max count of sealed block in epoch.
 const SEALED_BLOCK_IN_EPOCH: usize = 5;
+/// Max difference in timestamps of leader and witnesses.
+const TIME_TO_RECEIVE_BLOCK: u64 = 10 * 60;
```
rtsisyk: This is exactly what I want to do.

PR #342, opened 2019-01-17
Decouple Wallet from NodeService

rtsisyk:
This patch is on the top of #338. Please review #338 first.


rtsisyk:
Re-based. Please review,

PR #352, opened 2019-01-21
Completely reworked network stack, based on stock `rust-libp2p v0.2`

### LINKED ISSUE #322: Update to libp2p 0.1

echupriyanov:
* rewritten network stack
* defined network traits (`BroadcastNetwork` and `UnicastNetwork`)
* disabled `dial` command (never was used in real code, will add later, if needed)

Compiles cleanly, all tests are passing ok

Closes #322 

Don't merge yet, will have to squash before merge


network/protos/ncp_proto.proto
```diff
@@ -3,10 +3,8 @@ package ncp.pb;
 
 message Message {
 	enum MessageType {
-		PING = 0;
```
vldm: should we mention somewhere that now we have no heartbeat?

network/src/libp2p_network/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+        my_addresses.append(&mut multiaddresses);
+    }
+
+    debug!("My adverised addresses: {:#?}", my_addresses);
+    peerstore.add_local_external_addrs(my_addresses.into_iter());
+    peerstore
+}
+
+// Encode unicast message
+fn encode_unicast(from: secure::PublicKey, to: secure::PublicKey, data: Vec<u8>) -> Vec<u8> {
```
vldm: We probably need to move protobuf converting code from node to separate module. Because we used it highly intensive outside of node.

network/src/ncp/handler.rs
```diff
@@ -21,110 +21,238 @@
...
+        self.substreams.push(SubstreamState::WaitingInput(protocol))
+    }
+
+    fn inject_fully_negotiated_outbound(
+        &mut self,
+        protocol: <Self::OutboundProtocol as OutboundUpgrade<TSubstream>>::Output,
+        message: Self::OutboundOpenInfo,
+    ) {
+        if self.shutting_down {
+            return ();
```
vldm: You probably can omit `()` when returning unit type.

```suggestion
            return;
```

network/Cargo.toml
```diff
@@ -1,37 +1,32 @@
...
 unsigned-varint = "0.2"
-
-libp2p = { git = "https://github.com/stegos/rust-libp2p", rev = "stegos-0.0.1" }
-
-[dev-dependencies]
-simple_logger = "1.0"
-env_logger = "0.6"
+void = "1.0"
+libp2p = "0.2"
+# libp2p = { git = "https://github.com/stegos/rust-libp2p.git", rev = "stegos" }
```
vldm: Probably we should remove this line?

network/src/ncp/handler.rs
```diff
@@ -21,110 +21,238 @@
...
+    /// The active substreams.
+    // TODO: add a limit to the number of allowed substreams
+    substreams: Vec<SubstreamState<TSubstream>>,
+
+    /// Queue of values that we want to send to the remote.
+    send_queue: SmallVec<[NcpMessage; 16]>,
+}
+
+/// State of an active substream, opened either by us or by the remote.
+enum SubstreamState<TSubstream>
```
vldm: It would be good if we could write some small state-transition diagram. It could be ASCII graphicaly, or just doc text with info about transitions.

vldm:
LGTM, some nitpicks about docs, and comments removal only.

network/protos/ncp_proto.proto
echupriyanov: This particular type was never used in older code, so I just dropped it.
Heartbeat was a separate protocol on top of `floodsub`, which also was never used, so I dropped it too.

Not sure we should mention this.

network/src/libp2p_network/mod.rs
echupriyanov: This particular protobuf encodings are not used in any other parts of our code, so let's keep it closer to the place where it is used.

network/src/ncp/handler.rs
echupriyanov: This is a copy-paste from floodsub code. I'm fine with both variants.

network/Cargo.toml
echupriyanov: Yes, make sense.

network/src/ncp/handler.rs
echupriyanov: Let's create an issue for this and proceed further on

network/protos/ncp_proto.proto
rtsisyk: Let's remove all unused code. We don't have time to maintain dead code.

PR #357, opened 2019-01-22
Implement blocking of staked funds for 1 month

rtsisyk:


rtsisyk:
Fixed tests and re-based. Please review.

node/src/lib.rs
```diff
@@ -864,65 +832,87 @@ where
...
 
-            // Check that transaction's inputs are not used yet.
-            for tx_input_hash in &tx.body.txins {
-                if !inputs_hashes.insert(tx_input_hash.clone()) {
-                    let e = BlockchainError::MissingUTXO(tx_input_hash.clone());
+                // Check transaction's signature, monetary balance, fee and others
+                debug_assert!(tx.validate(&tx_inputs).is_ok()); // checked when added to mempool
+
+                Ok((tx_inputs, tx_outputs_hashes))
+            }() {
```
vldm: Could you split evaluation of the closure and match to different lines, because at this point it easilu to forget about match

node/src/lib.rs
```diff
@@ -671,63 +679,14 @@ where
         NodeService::flush_consensus_messages(consensus, &mut self.broker)
     }
 
-    /// Returns new active nodes list.
-    fn active_stakers(&self) -> Vec<(SecurePublicKey, i64)> {
```
vldm: It would good if this method would be left in some form at place, later we can add some logic to filter stakers that was inactive for a big period of time?

node/src/escrow.rs
```diff
@@ -0,0 +1,195 @@
...
+            validator_pkey: validator_pkey.clone(),
+            output_hash: hash_min,
+        };
+        let key_max = EscrowKey {
+            validator_pkey: validator_pkey.clone(),
+            output_hash: hash_max,
+        };
+
+        let mut stake: i64 = 0;
+        for (key, value) in self.escrow.range((Included(&key_min), Included(&key_max))) {
```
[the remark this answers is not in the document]
rtsisyk: I'll try.

node/src/lib.rs

node/src/lib.rs
rtsisyk: I wanted to move all this logic into Escrow.

rtsisyk:
Please don't merge yet. I plan to rework validation first and then re-base this patch.

rtsisyk:
Let's with with #372 first.

PR #367, opened 2019-01-23
Basic transaction pool Implementation.

vldm:
Introduce basic transaction pool implementation.
Without real work with transaction.
Without real facilitator choosing algorithm.
Without real mixing.

But with real handling network event.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -37,6 +37,7 @@ use stegos_crypto::pbc::secure;
 use stegos_keychain::KeyChain;
 use stegos_network::NetworkProvider;
 use stegos_node::*;
+use stegos_txpool::TransactionPool;
 use stegos_wallet::{Wallet, WalletNotification};
 
 // ----------------------------------------------------------------
@@ -57,8 +58,9 @@ where
         keys: &KeyChain,
         broker: Network,
         node: Node<Network>,
+        txpool: TransactionPool,
     ) -> Result<impl Future<Item = (), Error = ()>, Error> {
-        ConsoleService::new(keys, broker, node)
+        ConsoleService::new(keys, broker, node, txpool)
     }
 }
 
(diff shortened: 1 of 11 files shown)


blockchain/src/block.rs
```diff
@@ -97,10 +97,11 @@ pub struct KeyBlockHeader {
     /// Leader public key.
     pub leader: SecurePublicKey,
 
+    /// Facilitator of Transaction Pool.
+    pub facilitator: SecurePublicKey,
```
rtsisyk: Please include `facilitator` into Hashable for KeyBlockHeader.

src/console.rs
```diff
@@ -109,21 +117,26 @@ where
         keys: &KeyChain,
         broker: Network,
         node: Node<Network>,
+        txpool: TransactionPoolManager,
```
rtsisyk: Are you sure that you need TxPool intergation with Console?

txpool/protos/generated.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+package protobuf.pb;
+
+message SecurePublicKey {
```
rtsisyk: I think that this approach with re-implementing protobuf encoding is flaky. You will need encoding of Transaction object. In order to implement it, about 80% of  node/protos should be copy-pasted. 

Let's extract encoding of points, UTXOs, blocks and transactions from `stegos_node` into `stegos_blockchain` or a separate crate. We will need protobuf for persistence too. Probably `stegos_blockchain` is appropriate place, but I'm not sure.

txpool/src/api.rs
```diff
@@ -0,0 +1,63 @@
...
+    ChangeFacilitator(secure::PublicKey),
+}
+
+//TODO: Pool should return some commulative transaction.
+#[derive(Debug, Eq, PartialEq)]
+pub enum PoolFeedback {
+    Ping,
+}
+
+pub struct TransactionPoolManager {
```
rtsisyk: Typical convention here was `TransactionPool` for handler (API) and `TransactionPoolService` for Future.

txpool/src/protos/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+
+pub mod generated;
+pub mod messages;
+
+use stegos_crypto::pbc::secure;
+
+use failure::Error;
+use protobuf::Message as ProtobufMessage;
+
+pub trait StructIntoProto {
```
rtsisyk: Feel free to apply this refactoring to node/protos/mod.rs (by a separate patch).

src/console.rs
vldm: We need to somehow test it for now

Issue #304, opened 2019-01-11
Multiple transactions in a row


rtsisyk:
A meta ticket to track all issues need to be fixed in order to allow multiple successive transactions to be created in the single round.

**Wallet**:

- [ ] [Research how wallet can identify successful transaction](https://github.com/stegos/stegos/issues/326)
- [x] [Decouple wallet implementation from NodeService](https://github.com/stegos/stegos/issues/301)
- [ ] [Allow wallet to create multiple transactions in a row](https://github.com/stegos/stegos/issues/302)

**Mempool**:

- [ ] Add a queue for orphan transactions.
- [x] [Decouple mempool implementation from NodeService](https://github.com/stegos/stegos/issues/299)
- [x] [Prune transactions from mempool by UTXO ID](https://github.com/stegos/stegos/issues/300)
- [ ] [Allow transactions to use UTXOs from mempool transactions](https://github.com/stegos/stegos/issues/217)
- [ ] (Optional) [Order transactions in mempool](https://github.com/stegos/stegos/issues/303)

rtsisyk:
Not actual anymore.

Issue #299, opened 2019-01-11
Decouple mempool implementation from NodeService


rtsisyk:
Mempool is getting more complicated. All mempool-related logic should be extracted from node/src/lib.rs into a separate module:

- [x] Transactions queuing
- [x] Monetary block size estimation
- [x] Monetary block formation
- [x] Transactions prunning on a new sealed block

PR #372, opened 2019-01-24
Refactor transaction validation

### LINKED ISSUE #356: Remove invalid transactions from the mempool
Invalid transaction must be pulled out of the mempool. Currently they are kept forever.

### LINKED ISSUE #371: Remove double-checking for transactions accepted into mempool
Currently transactions are validated twice: on receiving and then a new sealed block is created. The second check is redundant. Transaction must be fully validated before accepting into mempool.

rtsisyk:
Before this patch transaction was validated twice:
    
1. In handle_transaction() before accepting into the mempool.
2. In process_mempool() before creating a new block.
    
Initial idea was that transaction can be received out of order,
therefore handle_transaction() should only perform basic checks,
whereas process_mempool() should peform the full check of inputs.
    
Actually, we can't check anything without having resolved inputs.
Even signature can't be checked without inputs. For this reason
transactions from the future can't be checked at all. Let's ignore
out-of-order transactions for now and focus on security.
    
This patch integrates all checks into handle_transaction().
Mempool now contains only fully validated transactions. Nothing can
fail during creation of a block. For security reason created block is
double-checked before sending it to the network.
    
Needed for #318
Needed for #299
Needed for #304
Closes #356
Closes #371

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -45,6 +45,7 @@ use protobuf;
 use protobuf::Message;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
+use std::collections::HashMap;
 use std::time::{Duration, Instant};
 use stegos_blockchain::*;
 use stegos_config::*;
@@ -248,6 +249,8 @@ struct NodeService<Network> {
 
     /// Memory pool of pending transactions.
     mempool: Mempool,
+    mempool_inputs: HashMap<Hash, Hash>,
+    mempool_outputs: HashMap<Hash, Hash>,
     /// Proof-of-stake consensus.
     consensus: Option<BlockConsensus>,
     /// A timestamp when the last sealed block was received.
@@ -285,6 +288,8 @@ where
         let vrf_system = TicketsSystem::new(WITNESSES_MAX, 0, 0, keys.cosi_pkey, keys.cosi_skey);
 
         let mempool = Mempool::new();
+        let mempool_inputs = HashMap::new();
+        let mempool_outputs = HashMap::new();
         let consensus = None;
         let last_block_timestamp = Instant::now();
 
@@ -348,6 +353,8 @@ where
             stakes,
             validators,
             mempool,
+            mempool_inputs,
+            mempool_outputs,
             consensus,
             last_block_timestamp,
             broker,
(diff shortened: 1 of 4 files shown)


rtsisyk:
Re-based.

dbmcclain:
You could actually check the UTXO Bulletproofs ahead of time, and mark the transactions that had been checked in this manner, so that you don't have to repeat that validation.

Validating Bulletproofs is no longer a serious issue for us. But when we began last year, BP validation was a serious time overhead, and so we did check BP's on first encounter with new transactions.

node/src/lib.rs
```diff
@@ -417,6 +424,23 @@ where
             tx.body.fee
         );
 
+        //
+        // Validation checklist:
```
dbmcclain: TX hash could possibly form a hash collision with other TX in mempool, and that wouldn't be invalid, though unlikely. The hash of the TX merely needs to validate against the TX Signature, and that can be checked immediately. It is merely a check to be sure TX has not been mutated.

Signature validation on the TX depends also on knowing the public key to be used for the whole TX, and so validation might not be possible until all TXIN are visible. But TX Hash is only one component of the validation. We use H(K, P, H(TX)) in the signature check. So H(TX) collision is not a problem here.

Sorry, there are two different interpretations of TX Hash. The one used in signature validation is the hash of everything in TX except the signature. But for database keying, you might use the hash of the entire TX, including signature, to form a lookup key. A hash collision in either of those, while unlikely, would still not be invalid. You can't blame a TX producer because he causes minor problems with your database scheme. What he presents must merely be shown to be valid.

rtsisyk:
> You could actually check the UTXO Bulletproofs ahead of time, and mark the transactions that had been checked in this manner, so that you don't have to repeat that validation.

Yes, exactly. That is the main idea of this patch. I validate entire transaction when it received from the network. No more validations are performed.

> Validating Bulletproofs is no longer a serious issue for us. But when we began last year, BP validation was a serious time overhead, and so we did check BP's on first encounter with new transactions.

We just enabled optimizations in cargo.toml even for Debug mode ;)

node/src/lib.rs
rtsisyk: BTW, `H(TX)` is also using during block proposal validation. I sent a new block + the list of `H(TX)`. Validators checks that transactions with `H(TX)` presents in their mempools, try to re-create the same block and check the signature. 

I can easily switch from `H(TX)` to `H(K, P, H(TX))` for database keying and validation. Any thoughts?

dbmcclain:
The H(K, P, H(TX)) is used in validating Schnorr signatures. They need to see something in the hash pre-image that is tied to the signature (K, u) itself, and to the supposed public key, P, of the signer.

But as you stated correctly (hopefull) hash collisions should be so rare that none of us wil ever see one in our lifetimes…. Use either hash as you wish…

- DM

rtsisyk:
Please review and approve or request changes. There are 3 pull requests in queue depending on this one.

@dbmcclain @echupriyanov @vldm

PR #386, opened 2019-01-28
Initial Release of DiceMix service

dbmcclain:
recode to one item per sheet of DiceMix matrix
cleanup and simplify with type declarations
add order comparisons to LEV32, Pt, and PublicKey
FLINT interface
add blame discovery to DiceMix

TODO: add some unit tests


dbmcclain:
Looks like we got bit again on the different order of cargo fmt between the linux and test builds

crypto/src/dicemix/solver_flint.cpp
```diff
@@ -0,0 +1,221 @@
+#include <vector>
```
rtsisyk: Is it a good idea to include these bindings into the primary project? Probably we can create a separate module, but I don't insist. 
@echupriyanov what do you think?

crypto/src/dicemix/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+sent any particular message. Anonymity is preserved when at least 3
+participants engage.
+*/
+
+// TODO: this stuff doesn't belong here in general crypto services
+// these are app-specific items.
+
+type TXIN = Hash;
+
+pub struct UTXO {
```
rtsisyk: I think that this part should extracted into a separate crate (e.g. stegos_txpool). I don't like the idea with copying UTXO structure from `stegos_blockchain` into `stegos_crypto`. I see that this structure is different. For example, `PaymentOutput` doesn't have `keying`.

crypto/src/dicemix/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+    pub id: Hash,
+    pub pkey: PublicKey,
+    pub bp: BulletProof,
+    pub keying: (PublicKey, PublicKey),
+    pub payload: [u8; 256],
+}
+
+fn serialize_utxo(_utxo: &UTXO) -> Vec<u8> {
+    let mut out = Vec::<u8>::new();
+    // TODO
```
rtsisyk: Let's remove all died code from this pull request and add it later when it would be needed.

crypto/src/dicemix/solver_flint.cpp
```diff
@@ -0,0 +1,221 @@
...
+#include <cstring>
+
+#include <flint/flint.h>
```
rtsisyk: We don't have `flint` library on the board. I have no idea why CI checks are green. 
BUILD.md and CI scripts should be updated for each new system library. Probably @echupriyanov can make a patch to integrate libflint properly.

crypto/src/dicemix/mod.rs
```diff
@@ -0,0 +1,644 @@
...
+    // of the hash over the first bytes of v.
+    let h = Hash::from_vector(v);
+
+    let p = h.bits();
+    for i in 0..NPREF {
+        v[i] = p[i];
+    }
+}
+
+fn split_message(msg: &[u8], max_cols: Option<usize>) -> DcRow {
```
rtsisyk: Such low-level code requires unit tests. Otherwise nobody can prove that this code works properly and doesn't crash.

dbmcclain:
I think this needs to be merged into dev, then incremental improvements made. Currently it is orthogonal to the rest of dev and should remain so until after unit tests.

One thing that comes to mind in thinking about how it will be used -- if blame discovery happens, then UTXO's need to be regenerated. That implies that the wallet needs to be intimately involved in ValueShuffle. It cannot simply construct transactions and submit them to VS group. 

Meanwhile, testing - Rust severely lacks infrastructure for interactive display of incremental constructs. A display scaffolding is needed here, along with non-muted I/O. Unit tests in Rust report nothing other than success/fail.

vldm:
@dbmcclain If test fails, Rust will write your "printlns" into terminal output.
`stderr` is not "muted" by default, so you can use log (but you need to init it in test itself).

rtsisyk:
> I think this needs to be merged into dev, then incremental improvements made. Currently it is orthogonal to the rest of dev and should remain so until after unit tests.

I agree that this code is orthogonal to the rest of dev. No one use it yet. No tests yet.
Of course, we can merge it now, but does it make sense? A good pull request should have some finished piece of work in order to be pushed. DiceMix itself is a huge bunch of work and we definitely want it, but is it finished? How to understand what it is finished and does work? I wonder how do you test it? 
 
> One thing that comes to mind in thinking about how it will be used -- if blame discovery happens, then UTXO's need to be regenerated. That implies that the wallet needs to be intimately involved in ValueShuffle. It cannot simply construct transactions and submit them to VS group.

I didn't expected that. Let me learn how ValueShuffle works in order to help with this problem. 

> Meanwhile, testing - Rust severely lacks infrastructure for interactive display of incremental constructs. A display scaffolding is needed here, along with non-muted I/O. Unit tests in Rust report nothing other than success/fail.

Try to enable logging in unit tests:

```
    #[test]
    fn basic() {
        use simple_logger;
        simple_logger::init_with_level(log::Level::Debug).unwrap_or_default();
   }
```

dbmcclain:
Not my problem (yet...) 

But when using VS, wallets will take a different path to constructing transactions. They will not actually construct a transaction, but rather go through the VS dance of sharing session keys with other participants, along with all their TXINS and individual "ownership signatures" on each TXIN. Then they will construct their UTXOs (up to 5 of them) and the gamma_adj value, and anonymously share this info with other participants. 

If all goes well, then a super-transaction from all participants results, wallets are asked to collectively sign the super transaction, and then it is forwarded to the Cosi validators. 

If anything goes wrong, the list of participants is trimmed to exclude the trouble maker, then wallets must regenerate their UTXOs and gamma_adj value with fresh randomness and re-submit through anonymous sharing with the group.

The iterations continue until success or total failure. So there is a lot of back and forth between wallet and VS group in either case.

dbmcclain:
looks like the Linux Test pipeline has issues with the std::dbg::dbg!() macro...

rtsisyk:
I added `#[ignore]` to tst_50_50 because it was so slow. Please use `--ignored` to run it:

```
cargo test -p stegos_crypto dicemix::tests::tst_50_50 -- --ignored
```

rtsisyk:
Re-based and merged manually.

PR #405, opened 2019-01-31
Added `protocol id` field to unicast messages.

### LINKED ISSUE #402: Update Unicast API for "simulated" implementation
* Introduce `protocol_id` field to enable routing messages to specific handlers
* ~Add Error message type to messages passed to handlers to relay transport errors~ (team decided we don't need feedback for unicast communications)

echupriyanov:
Allows demultiplexing unicast messages to corresponding subsystems

Closes #402


network/src/libp2p_network/mod.rs
```diff
@@ -421,14 +453,16 @@ fn encode_unicast(from: secure::PublicKey, to: secure::PublicKey, data: Vec<u8>)
...
 fn decode_unicast(
     input: Vec<u8>,
-) -> Result<(secure::PublicKey, secure::PublicKey, Vec<u8>), Error> {
+) -> Result<(secure::PublicKey, secure::PublicKey, String, Vec<u8>), Error> {
     let mut msg: unicast_proto::Message = protobuf::parse_from_bytes(&input)?;
 
     let from = secure::PublicKey::try_from_bytes(&msg.take_from().to_vec())?;
     let to = secure::PublicKey::try_from_bytes(&msg.take_to().to_vec())?;
+    let protocol_id_bytes = &msg.get_protocol_id();
+    let protocol_id = String::from_utf8_lossy(protocol_id_bytes).to_string();
```
vldm: Why not using `from_utf8` and return error if some invalid protocol id was provided?

echupriyanov: `protocol_id` functions just as a tag to pass received message to corresponding subsystem, so in case of invalid `protocol_id` there will be no subscribers, and message will be just dropped on the floor.

Anyway this implementation is just a temporal solution, until we have normal unicast.
I'll think about handling invalid `protocol_id` there.

rtsisyk: There are two options:
- Force `utf-8`.
- Use `protocol_id` as opaque value. In this case utf-8 is not need to be checked. For this case OsString or just Vec<u8> should be used.

network/protos/unicast_proto.proto
```diff
@@ -5,4 +5,5 @@ message Message {
     required bytes from = 1;
     required bytes to = 2;
     required bytes data = 3;
+    required bytes protocol_id = 4;
```
rtsisyk: Please swap `data` and `protocol_id`.

network/src/lib.rs
```diff
@@ -47,14 +47,30 @@ pub trait NetworkProvider {
...
     /// Returns a list of all the known peers in the topology.
     fn peers(&self) -> Vec<&PeerId>;
 }
+
+/// Placeholder for NetworkError definitions
+/// TODO: Add/implement real errors
+#[derive(Clone, Debug, Fail, PartialEq, Eq)]
+pub enum NetworkError {
+    #[fail(display = "Generic network error talking to node: {:#?}", _0)]
+    GenericError(secure::PublicKey),
```
rtsisyk: Looks strange.

echupriyanov: Could you elaborate?

network/protos/unicast_proto.proto
echupriyanov: Ok

network/src/lib.rs
rtsisyk: What is "GenericError"? Why do you need "GenericError"?

echupriyanov: In the comments above - "Reserved placeholder for future Errors".
When actual Errors will be refined, they will appear here.

Issue #320, opened 2019-01-14
Initial blockchain state transfer


rtsisyk:
Request initial blockchain state from a source node and apply this state locally.

PR #410, opened 2019-02-04
Initial bootstrap implementation.

vldm:
Closes: #319 
Closes: #320 

- Implemented basic network bootstrap.
- ConsensusMessages in one epoch from ALL future height now are silently added into queue.

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -245,6 +245,20 @@ impl Blockchain {
         return None;
     }
 
+    /// Returns blocks history starting from block_hash, limited by count.
+    pub fn blocks_range(&self, starting_hash: &Hash, count: u64) -> Option<&[Block]> {
+        if let Some(&block_id) = self.block_by_hash.get(starting_hash) {
+            let block_id = block_id + 1;
+            let len = self.blocks.len();
+            let end_range = usize::min(len, block_id + count as usize);
+
+            if block_id < end_range {
+                return Some(&self.blocks[block_id..end_range]);
+            }
+        }
+        return None;
+    }
+
     /// Return all blocks.
     pub fn blocks(&self) -> &[Block] {
         self.blocks.as_slice()
--- a/consensus/src/state.rs
+++ b/consensus/src/state.rs
@@ -278,9 +278,6 @@ impl<Request: Hashable + Clone + Debug, Proof: Hashable + Clone + Debug> Consens
             return Err(ConsensusError::UnknownMessagePeer(msg.pkey));
         }
 
-        // Validate signature and content.
-        msg.validate()?;
-
         // Check round.
         if msg.height < self.height {
             debug!(
@@ -291,24 +288,15 @@ impl<Request: Hashable + Clone + Debug, Proof: Hashable + Clone + Debug> Consens
             );
             // Discard this message.
             return Ok(());
-        } else if msg.height == self.height + 1 {
+        } else if msg.height > self.height {
             debug!(
-                "{}({}): message from the next round: msg={:?}",
+                "{}({}): message from the future: msg={:?}",
                 self.state.name(),
                 self.height,
                 &msg
             );
             // Queue the message for future processing.
             self.inbox.push(msg);
-            return Ok(());
-        } else if msg.height > self.height + 1 {
-            warn!(
-                "{}({}): message from the future: msg={:?}",
-                self.state.name(),
-                self.height,
-                &msg
-            );
-            // Discard this message.
             return Ok(());
         }
         assert_eq!(msg.height, self.height);
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -24,6 +24,7 @@
 mod consensus;
 mod election;
 mod error;
+mod loader;
 mod mempool;
 pub mod protos;
 mod tickets;
@@ -34,6 +35,7 @@ use bitvector::BitVector;
 
 use crate::election::ConsensusGroup;
 use crate::error::*;
+use crate::loader::ChainLoader;
 pub use crate::tickets::{TicketsSystem, VRFTicket};
 use chrono::Utc;
 use failure::{ensure, Error};
@@ -57,6 +59,7 @@ use stegos_crypto::pbc::secure::PublicKey as SecurePublicKey;
 use stegos_crypto::pbc::secure::Signature as SecureSignature;
 use stegos_keychain::KeyChain;
 use stegos_network::Network;
+use stegos_network::UnicastMessage;
 use stegos_serialization::traits::ProtoConvert;
 use tokio_timer::Interval;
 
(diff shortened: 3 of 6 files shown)


node/src/chain_loader.rs
```diff
@@ -0,0 +1,228 @@
+//
```
rtsisyk: m.b. just `loader.rs` or even better `replication.rs`?

node/src/chain_loader.rs
```diff
@@ -0,0 +1,228 @@
...
+use stegos_blockchain::Block;
+use stegos_crypto::hash::{Hash, Hashable, Hasher};
+use stegos_crypto::pbc::secure;
+use stegos_serialization::traits::ProtoConvert;
+
+use crate::consensus::SealedBlockMessage;
+use crate::NodeService;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub struct BlocksRange {
```
rtsisyk: BlockRange (without `s`)

node/src/lib.rs
```diff
@@ -1338,9 +1360,14 @@ impl Future for NodeService {
...
                         NodeMessage::Transaction(msg) => self.handle_transaction(msg),
                         NodeMessage::Consensus(msg) => self.handle_consensus_message(msg),
-                        NodeMessage::SealedBlock(msg) => self.handle_sealed_block(msg),
+                        NodeMessage::SealedBlock(msg) => SealedBlockMessage::from_buffer(&msg)
+                            .and_then(|msg| self.handle_sealed_block(msg)),
                         NodeMessage::ConsensusTimer(_now) => self.handle_consensus_timer(),
                         NodeMessage::VRFMessage(msg) => self.handle_vrf_message(msg),
+                        NodeMessage::ChainLoaderMessage(msg) => {
+                            chain_loader::ChainLoaderMessage::from_buffer(&msg.data)
+                                .and_then(|data| self.handle_chain_loader_message(msg.from, data))
```
rtsisyk: Please make a patch on the top of `dev` to change all similar places to use this approach.

node/src/chain_loader.rs
rtsisyk: BTW, the only use of this structure is `RequestBlocks`. Does it make sense to create a separate structure? Let's integrate it into `RequestBlocks` to simplify things.

node/src/chain_loader.rs
```diff
@@ -0,0 +1,228 @@
...
+    fn hash(&self, state: &mut Hasher) {
+        self.range.hash(state);
+    }
+}
+
+impl RequestBlocks {
+    pub fn new(current_hash: Hash, block_hash: Hash) -> RequestBlocks {
+        let range = BlocksRange {
+            start_block: current_hash,
+            end_block: block_hash,
```
rtsisyk: This limit is never used in the code.

consensus/src/state.rs
```diff
@@ -278,9 +278,6 @@ impl<Request: Hashable + Clone + Debug, Proof: Hashable + Clone + Debug> Consens
             return Err(ConsensusError::UnknownMessagePeer(msg.pkey));
         }
 
-        // Validate signature and content.
-        msg.validate()?;
```
rtsisyk: Is it intentional?

node/src/lib.rs
```diff
@@ -674,14 +674,45 @@ impl NodeService {
...
-
+    fn handle_sealed_block(&mut self, msg: SealedBlockMessage) -> Result<(), Error> {
         // Check signature and content.
         msg.validate()?;
 
+        let ref block = msg.block;
+        let block_hash = Hash::digest(block);
+
+        if let Block::MonetaryBlock(_) = block {
+            // TODO: Should this check exist? We can send block as response, and it would be validated without this check.
```
rtsisyk: Which one? `if let Block::MonetaryBlock(_) = block`?
Monetary block should be signed by the epoch leader. KeyBlock should be signed by a leader elected by VRF. It is a good question to check VRF leader.

node/src/lib.rs
```diff
@@ -674,14 +674,45 @@ impl NodeService {
...
+                return Err(NodeError::SealedBlockFromNonLeader(
+                    block_hash,
+                    self.chain.leader.clone(),
+                    msg.pkey,
+                )
+                .into());
+            }
+        }
+
+        {
```
rtsisyk: Probably these brackets are not more needed in Rust 2018.

node/src/chain_loader.rs
```diff
@@ -0,0 +1,228 @@
...
+            ChainLoaderMessage::Response(r) => {
+                "response".hash(state);
+                r.hash(state)
+            }
+        }
+    }
+}
+
+/// Limit of blocks to download starting from current known blockchain state.
+const ORPHANS_BLOCK_LIMIT: u64 = 1000;
```
rtsisyk: Too many. Master can be easily DDoSed by such requests. Probably good value is 10 or 100..

node/src/chain_loader.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl ChainLoader {
+    pub fn new() -> Self {
+        ChainLoader {
+            blocks_queue: Vec::new(),
+        }
+    }
+}
+
+//TODO: Download blocks from different participant.
```
rtsisyk: Please don't forget to open tickets for this problems after finishing with this pull request.

blockchain/src/blockchain.rs
```diff
@@ -245,6 +245,20 @@ impl Blockchain {
         return None;
     }
 
+    /// Returns blocks history starting from block_hash, limited by count.
+    pub fn blocks_range(&self, starting_hash: &Hash, count: u64) -> Option<&[Block]> {
```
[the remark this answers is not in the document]
vldm: @rtsisyk Yes, but we already have blocks() method that return &[Block] and many code related to it len and context. We can rewrite this code to return Iterator, or something more specific. Do you want me to return iterator from rocksdb in this pullrequest, or we can deal with it after second version of persistance will be implemented?

node/src/chain_loader.rs
vldm: @rtsisyk I started with implementing more complex bootstrap system, in which this structure is used as part of responses too.
For now i can simplify it, or if we want to implement more complex bootstrap system in near future, we can keep this structure for future refactor?

consensus/src/state.rs
vldm: Yes, `feed_message` is internal method that handles internal (created by this node) and external messages. I refactored queue a bit, and found that validate could be called more then once for messages in queue. So i decide to move verify into `handle_consensus_message`

node/src/lib.rs
vldm: @rtsisyk Okay, i get your point.

But this comment was about difference between Block and Sealed block message. We only keep signature of the leader in the sealed block message. But in the storage we only keep Block without leader signature, so when we sending blocks as BlocksResponse, we omit leader signature.

So do we really need to check leader signature? Its looks like redurant.

PR #419, opened 2019-02-07
Fix unicast sending to itself + console send support

### LINKED ISSUE #417: Subscribe ConsoleService to unicast message with `console` protocol_id

echupriyanov:
* Handle case of unicast messgaes to self
* Subscribe console to unicast messages
  * subscribe to unicast stream with protocol id = `console`
  * display received messages

Closes #417, #418

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -34,7 +34,7 @@ use std::thread;
 use stegos_crypto::curve1174::cpt::PublicKey;
 use stegos_crypto::pbc::secure;
 use stegos_keychain::KeyChain;
-use stegos_network::Network;
+use stegos_network::{Network, UnicastMessage};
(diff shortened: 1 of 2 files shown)


src/console.rs
```diff
@@ -457,6 +463,18 @@ impl Future for ConsoleService {
             }
         }
 
+        loop {
+            // Process unicast messages
+            match self.unicast_rx.poll() {
+                Ok(Async::Ready(Some(msg))) => {
+                    info!("Received unicast message. From: {}, Data: {}", msg.from, String::from_utf8_lossy(&msg.data));
```
rtsisyk: I use the following convention for messages:

Message text: arg1=xxx, arg2=xxx

PR #465, opened 2019-02-14
Extract validation code from node/src/lib.rs

rtsisyk:
Extract validation code from node/src/lib.rs
    
No semantic changes. This patch is pure refactoring.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -30,6 +30,7 @@ pub mod protos;
 #[cfg(test)]
 mod test;
 mod tickets;
+mod validation;
 
 use crate::consensus::*;
 use crate::mempool::Mempool;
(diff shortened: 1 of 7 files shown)


node/src/lib.rs
```diff
@@ -586,109 +509,41 @@ impl NodeService {
...
+        if previous_hash != header.previous {
+            let block_hash = Hash::digest(block);
+            debug!(
+                "Received orphan block: hash={}, expected_previous={}, got_previous={}",
+                &block_hash, &previous_hash, &header.previous
+            );
+            return self.on_orphan_block(msg);
+        }
+
+        if header.epoch != self.chain.epoch {
```
vldm: Why you need this check?
If block linked to previous (and it is valid), this block is a part of blockchain, even if it from future epoch, have i missed something?

rtsisyk: Probably yes. I just move this check to upper level from some helper. I will fix it during adding tests.

Issue #483, opened 2019-02-18
Check genesis block during recovery


rtsisyk:
Data on the disk must match genesis block

PR #489, opened 2019-02-18
Check genesis block during recovery.

vldm:
Closes: #483

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -361,6 +361,19 @@ impl NodeService {
         if self.chain.blocks().len() > 0 {
             //rather recover state
             self.recover_state();
+            for ((height, genesis), chain) in genesis.iter().enumerate().zip(self.chain.blocks()) {
+                let genesis_hash = Hash::digest(genesis);
+                let chain_hash = Hash::digest(chain);
+                if genesis_hash != chain_hash {
+                    error!(
+                        "Found a saved chain that is not compatible to our genesis at height = {}, \
+                         genesis_block = {:?}, database_block = {:?}",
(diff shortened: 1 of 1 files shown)


node/src/lib.rs
```diff
@@ -361,6 +361,13 @@ impl NodeService {
         if self.chain.blocks().len() > 0 {
             //rather recover state
             self.recover_state();
+            for (genesis, chain) in genesis.iter().zip(self.chain.blocks()) {
+                if Hash::digest(genesis) != Hash::digest(chain) {
+                    warn!("Found a saved chain that is not compatible to our genesis.");
```
rtsisyk: Must not continue in this case. Please print both genesis blocks like in main().

vldm: @rtsisyk but then if some user has lost his genesis, then it being inpossible to load data. Are you sure that we need this behaviour?

rtsisyk: Who do you mean by "lost his genesis"?

vldm: @rtsisyk For example use binary with wrong genesis.

node/src/lib.rs
```diff
@@ -361,6 +361,19 @@ impl NodeService {
         if self.chain.blocks().len() > 0 {
             //rather recover state
             self.recover_state();
+            for ((height, genesis), chain) in genesis.iter().enumerate().zip(self.chain.blocks()) {
+                let genesis_hash = Hash::digest(genesis);
+                let chain_hash = Hash::digest(chain);
+                if genesis_hash != chain_hash {
+                    warn!(
```
rtsisyk: error, not warn!

PR #492, opened 2019-02-18
Fix bootstrap with pruned blocks.

vldm:


rtsisyk:
Please check manually that it works.

PR #496, opened 2019-02-19
Resolve seed nodes from SRV and A records.

echupriyanov:
If we are running on testnet, first lookup SRV records for value of config.network.seed_pool
Take port and hostname(s) from SRV records and than lookup hostname(s) to get IP addresses.


src/stegos.rs
```diff
@@ -141,6 +145,26 @@ fn initialize_genesis(cfg: &config::Config) -> Result<Vec<Block>, Error> {
     Ok(blocks)
 }
 
+fn resolve_pool(cfg: &mut config::Config) -> Result<(), Error> {
+    if cfg.network.seed_pool != "" {
```
rtsisyk: Just return in this case. Don't wrap the main case into a block.

src/stegos.rs
```diff
@@ -79,7 +80,10 @@ fn load_configuration(args: &ArgMatches<'_>) -> Result<config::Config, Error> {
     if let Some(chain) = args.value_of("chain") {
         cfg.general.chain = chain.to_string();
     }
-
+    // UGLY! UGLY! UGLY!
```
rtsisyk: Please add a proper comment.

PR #516, opened 2019-02-20
WIP: Initial consensus tests.

vldm:


blockchain/src/block.rs
```diff
@@ -88,7 +101,7 @@ impl Hashable for BaseBlockHeader {
 }
 
 /// Header for Key Blocks.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
```
rtsisyk: I don't think that comparing blocks byte by byte is a good idea.

rtsisyk:
Since we have plans to change consensus, I'm moving the PR to the next sprint.

rtsisyk:
We work on a new version of VGF && Consensus algorithm. Let's add test after finishing implementation.

Issue #521, opened 2019-02-21
Chain loading didn't take into account limit.


vldm:
Regression of https://github.com/stegos/stegos/pull/492

Chain loading should take into account limit.

PR #522, opened 2019-02-21
Fix chain loading limit.

vldm:
Closes: https://github.com/stegos/stegos/issues/521


vldm:
Need to add regresiion test.

Issue #440, opened 2019-02-11
Query for blocks you don't have


vldm:
We should use request-response, and reduce queueing in
1) Consensus messages.
2) Monetary block with transactions.

Queue is bad because it potential DoS.

PR #523, opened 2019-02-21
Request blocks from node that send future ticket.

vldm:
related to https://github.com/stegos/stegos/issues/440

### DIFF
--- a/node/src/loader.rs
+++ b/node/src/loader.rs
@@ -168,9 +168,14 @@ impl NodeService {
         return master;
     }
 
-    /// Request block history, from one of validators.
+    /// Request the block history from an random node.
     pub fn request_history(&mut self) -> Result<(), Error> {
         let master = self.choose_master();
+        self.request_history_from(master)
+    }
+
+    /// Request the block history from specified node.
+    pub fn request_history_from(&mut self, master: secure::PublicKey) -> Result<(), Error> {
         let last_hash = self.chain.last_block_hash();
         info!(
             "Downloading blocks: from={}, last_hash = {}",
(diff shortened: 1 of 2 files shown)


node/src/loader.rs
```diff
@@ -258,4 +258,38 @@ impl NodeService {
             ChainLoaderMessage::Response(r) => self.handle_response_blocks(pkey, r),
         }
     }
+
+    pub fn handle_message_from_future(
```
rtsisyk: The name is so generic... Which kind of message?

rtsisyk: Why handle_message_from_future() is in loader.rs, but called from tickets.rs?

vldm: @rtsisyk 
>Which kind of message?

VRF Ticket, or consensus message. All message that has field `height`, and height is greater than our height.

>Why handle_message_from_future() is in loader.rs, but called from tickets.rs?

Yes this is a kind of cross module dependencies.


For now i only see three sources of messages from future:
1) Sealed block from future height
2) Consensus message from future height
3) Vrf ticket from future height

Downloading by receiving 1st message was already implemented as basic node bootstrap, but we have next problems, with it:
1) We cant start load faster then in period of sealed block propagation.
2) if some node has restarted just before last block at epoch, it will skip the VRF ticketing process (because it will be at other lower height).

rtsisyk: There is no need to introduce cross-module dependencies. 
I refactored this patch to remove this dependency. Please review and merge if you are agree.

vldm:
@rtsisyk LGTM, but i can't approve my pullrequests.

PR #525, opened 2019-02-21
Improve badges in README.md

rtsisyk:


rtsisyk:
I think a link to website should be added too.

PR #527, opened 2019-02-22
Link libstdc++ statically

rtsisyk:
I haven't found solution for `libgcc_s`. `println!("cargo:rustc-link-lib=static=gcc_s")` doesn't work.
Let's try with shared libgcc_s first, it should work.


rtsisyk:
Doesn't help

PR #551, opened 2019-02-25
Implement Drop trait on secret keys

dbmcclain:
Zap secret key memory to zeros before deallocation

### DIFF
--- a/crypto/src/curve1174/cpt.rs
+++ b/crypto/src/curve1174/cpt.rs
@@ -25,6 +25,7 @@
 use super::*;
 use crate::CryptoError;
 
+use clear_on_drop::clear::Clear;
 use crypto::aes;
 use crypto::aes::KeySize::KeySize128;
 use crypto::aesni;
@@ -46,7 +47,7 @@ impl Pt {
     }
 
     /// Convert into raw bytes.
-    pub fn into_bytes(self) -> [u8; 32] {
+    pub fn to_bytes(&self) -> [u8; 32] {
         return self.0;
     }
 
@@ -65,7 +66,7 @@ impl Pt {
     /// Create from an uncompressed point.
     #[inline]
     pub fn compress(ept: ECp) -> Pt {
-        let bytes = ECp::into_bytes(ept);
+        let bytes = ECp::to_bytes(&ept);
         Pt(bytes)
     }
 
@@ -76,7 +77,7 @@ impl Pt {
     }
 
     /// Convert into hex string.
-    pub fn into_hex(self) -> String {
+    pub fn to_hex(&self) -> String {
         let v = Lev32(self.0);
         basic_nbr_str(&v.to_lev_u64())
     }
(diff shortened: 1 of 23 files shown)


dbmcclain:
Implementing a Drop trait on SecretKeys, also changes slightly the way they must be used. So I also had to recode various sections that use SecretKeys to perform with clone() of secret keys. Any items with Drop traits cannot implement or inherit the Copy methods.

crypto/src/curve1174/cpt.rs
```diff
@@ -160,13 +160,19 @@ impl SecretKey {
 
 impl fmt::Display for SecretKey {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "SKey({})", self.into_hex())
+        write!(f, "SKey({})", self.clone().into_hex())
```
rtsisyk: try &self.into_hex()

dbmcclain: To be honest, I don't have a clear picture of what Rust is doing behind the scenes either. But without the clone() you get a borrow checker violation. I think what the clone() is doing here is constructing an overt copy of the original secret key, which it can safely deconstruct on exit, using the Drop traits that I defined.

dbmcclain: I think the lifetime analysis in Rust is still too rudimentary to track whole program ownership and see safe locations for final deconstruction.

vldm: This is not part of "lifetime analysis". Method into hex has next signature `into_hex(self) -> String`. `into_hex` not really needs  taking self by value, we can modify it to `into_hex(&self)->String`

This mean that you should provide `self` by value, rather then borrowing it.

crypto/src/curve1174/u256.rs
```diff
@@ -48,6 +48,13 @@ impl U256 {
         U256([1, 0, 0, 0])
     }
 
+    pub fn zap(&mut self) {
+        let mut bytes = self.0;
```
vldm: This probably a broken implementation of zeroing. In this code you mooving self.0 into bytes, which internally create allocate a new space on stack, and then clear only that place.

Please use https://crates.io/crates/clear_on_drop 

Example of usage:
https://docs.rs/clear_on_drop/0.2.3/clear_on_drop/clear/index.html

node/src/tickets.rs
```diff
@@ -546,7 +546,7 @@ mod test {
         }
 
         let seed = mix(block_hash, view_change + 1);
-        let ticket = VRFTicket::new(seed, height, p, &s);
+        let ticket = VRFTicket::new(seed, height, p, &s.clone());
```
vldm: why you clone there? In this code we only need a reference of &SecretKey

crypto/src/curve1174/cpt.rs
rtsisyk: I'm fixing this problem right now.

crypto/src/curve1174/u256.rs
rtsisyk: Done

node/src/tickets.rs
rtsisyk: Fixed

PR #559, opened 2019-02-26
Implement super-transactions for ValueShuffle

### LINKED ISSUE #328: Super-transaction formation and broadcast

rtsisyk:
- Upgrade Transactions to support TXINs to multiple payee pubkeys,
  all belonging to one node who knows the corresponding secret keys.

- Added crypto primitive curve1174::cpt::sign_hash_with_kval() to support
  construction of Schnorr multi-signatures.

- Add operators + and += for SchnorrSig to support multi-signatures.

- Corrections to Schnorr multi-signature construction.

- Shorten keying hint in UTXO EncryptedPayload to use only
  hint = alpha * P, key = H(s * alpha * G).
  Saves 32 bytes (256 bits) from every UTXO

Closes #328

### DIFF
--- a/crypto/src/protos.rs
+++ b/crypto/src/protos.rs
@@ -24,7 +24,7 @@ use stegos_serialization::traits::*;
 
 use crate::bulletproofs::{BulletProof, DotProof, L2_NBASIS, LR};
 use crate::curve1174::cpt::Pt;
-use crate::curve1174::cpt::{EncryptedPayload, PublicKey, SchnorrSig};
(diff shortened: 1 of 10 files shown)


rtsisyk:
@dbmcclain I extracted this part from vs-outer-3. I think we can merge it, but please add some basic unit tests for Transaction::new_super_transaction(). Thanks!

crypto/src/protos.rs
```diff
@@ -95,6 +95,20 @@ impl ProtoConvert for Hash {
     }
 }
 
+impl ProtoConvert for SecretKey {
```
rtsisyk: Why do you need to send secret key?

PR #560, opened 2019-02-26
Add view_change to vrf ticket message.

vldm:
Closes: https://github.com/stegos/stegos/issues/533


node/src/tickets.rs
```diff
@@ -33,7 +33,7 @@ use crate::election::{self, ConsensusGroup, StakersGroup};
 use crate::NodeService;
 
 use failure::{Error, Fail};
-use log::{debug, info, trace, warn};
+use log::{debug, info, trace};
```
rtsisyk: I think it is more convenient to "use log::*"

PR #570, opened 2019-02-28
Test brute force

dbmcclain:
Example code in stegos_crypto to provide timing estimates on brute force recovery of amounts from Bulletproof Pedersen commitments. 

This serves as a template for future code that can help recipients recover correct amounts when they detect a cheating UTXO where the encrypted amount differs from what had been used in the Bulletproof.

My timing estimates show that 28 bits can be recovered in about 90 sec.

### DIFF
--- a/crypto/examples/bruteforce.rs
+++ b/crypto/examples/bruteforce.rs
@@ -1,0 +1,65 @@
+// bruteforce.rs - timing tests on brute force ECDLP
+//
+// Copyright (c) 2018 Stegos AG
+//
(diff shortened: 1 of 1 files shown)


vldm:
@dbmcclain There nice library https://github.com/bheisler/criterion.rs in rust, to provide statistically driven benches.

@dbmcclain Rust has more simplifyed copy of that https://doc.rust-lang.org/1.15.1/book/benchmark-tests.html

We could use it one of them to write benches.

Issue #394, opened 2019-01-30
Test Plan for Blockchain Database


rtsisyk:
**Encoding/Decoding:**

- [ ] Invalid protobuf packet
- [ ] Missing mandatory fields
- [ ] Duplicate witnesses

**Basic Processing:**

- [ ] Invalid version
- [ ] Invalid epoch 
- [ ] Invalid timestamp
- [ ] Invalid multisignature
- [ ] Bitmap is too large
- [ ] Multisignature has 2/3 
- [ ] Invalid SealedBlockMessage signature
- [ ] Sender of SealedBlockMessage is not leader
- [ ] Already registered / duplicate hash
- [ ] Missing previous block

**Key Block:**

- [ ] No witnesses (error)
- [ ] Leader is not witness (error)
- [ ] Genesis block validation

**Monetary Block:**

- [ ] Invalid monetary balance
- [ ] One input (PaymentUTXO), fee, no outputs (should work)
- [ ] One input (DataUTXO), no outputs (should work) 
- [ ] No inputs, one output (should work)
- [ ] No inputs, no outputs (should work)
- [ ] StakeUTXO as input
- [ ] StakeUTXO as output
- [ ] Mutated gamma
- [ ] Invalid input_range_hash
- [ ] Invalid output_range_hash
- [ ] Min/max of  monetary_adjustment
- [ ] Duplicate input inside block
- [ ] Duplicate output inside block
- [ ] Missing input
- [ ] Output hash collision in blockchain 
- [ ] Global monetary balance
- [ ] TX exists in mempool
- [ ] Re-created block has the same hash as proposed
- [ ] Genesis block validation
- [ ] Block with zero fee

Plus UTXO-specified tests from https://github.com/stegos/stegos/issues/391:

**PaymentUTXO:**

- [ ] Mutated recipient 
- [ ] Invalid bullet proofs
- [ ] Zero amount (should work)
- [ ] Negative amount 
- [ ] Truncated payload

**Stake UTXO:**
 
- [ ] Mutated recipient  
- [ ] Zero or negative amount  
- [ ] Amount is less than PAYLOAD_FEE (error)
- [ ] Truncated payload
- [ ] Stake locking

**Other**:

- [ ] Invalid serialized Merkle Tree
- [ ] Serialization of pruned Merkle Tree
- [ ] Blockchain::output_by_hash() for a missing utxo
- [ ] Blockchain::outputs_by_hashes()
- [ ] Blockchain::block_by_hash
[...]

PR #579, opened 2019-03-05
Refactor block validation code

### LINKED ISSUE #537: Add leader multisignature to BlockPropose message

rtsisyk:
- Merge key_block.validate(), node.validate_sealed_key_block() and
  validation checks from chain.register_key_block() into one method.
- Merge monetary_block.validate(), node.validate_sealed_monetary_block() and
  validation checks from chain.register_monegtary_block() into one method.
- Remove double validation of monetary blocks.
- Remove validation of blocks during recovery.
- Add leader's signature into block's proposals.
- Validate that block's proposals always have leader's signature.
- Fix a bug that loaded blocks from the disk were written back to the disk.

This patch disables tests for block.validate() code. These tests
will be ported to the new codebase by #394.

Closes #537
Needed for #394
Needed for #539

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -427,7 +427,7 @@ impl NodeService {
                         Hash::digest(&key_block)
                     );
                     let key_block2 = key_block.clone();
-                    self.chain.register_key_block(key_block)?;
+                    self.chain.push_key_block(key_block)?;
                     self.on_key_block_registered(&key_block2)?;
                 }
                 Block::MonetaryBlock(monetary_block) => {
@@ -436,13 +436,10 @@ impl NodeService {
                         self.chain.height() + 1,
                         Hash::digest(&monetary_block)
                     );
-                    monetary_block
-                        .validate(&[])
-                        .expect("monetary balance is ok");
                     let monetary_block2 = monetary_block.clone();
                     let (inputs, outputs) = self
                         .chain
-                        .register_monetary_block(monetary_block, current_timestamp)?;
+                        .push_monetary_block(monetary_block, current_timestamp)?;
                     self.on_monetary_block_registered(&monetary_block2, inputs, outputs)?;
                 }
             }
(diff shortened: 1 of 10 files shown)


node/src/lib.rs
```diff
@@ -786,7 +797,7 @@ impl NodeService {
             self.consensus = Some(consensus);
             let consensus = self.consensus.as_ref().unwrap();
             if consensus.is_leader() {
-                self.create_new_epoch(self.chain.facilitator)?;
+                self.create_new_epoch(self.chain.leader)?;
```
vldm: Why you changed this?

node/src/mempool.rs
```diff
@@ -342,8 +342,10 @@ mod test {
             let input_hash = Hash::digest(input);
             inputs.insert(input_hash, input.clone());
         }
-        let inputs: Vec<Output> = inputs.values().cloned().collect();
-        block.validate(&inputs).expect("block is valid");
+        // TODO: validate?
+        //let inputs: Vec<Output> = inputs.values().cloned().collect();
+        //block.validate(&inputs).expect("block is valid");
+        drop(block);
```
vldm: This change looks strange, can you clarify what is your plan on it?

node/src/lib.rs
rtsisyk: Because it was a bug. The key block must be created by leader, not by facilitator.

node/src/mempool.rs
rtsisyk: I can revert block.validate() just for tests.

node/src/lib.rs
vldm: @rtsisyk Keyblock is created using current consensus.
But facilitator is received as argument, to add it to keyblock.

node/src/mempool.rs
vldm: Ok, i missed that it was edit in test, its ok, just don't forget to fix it.

node/src/lib.rs
rtsisyk: Fixed

rtsisyk:
Pushed an updated version.

PR #580, opened 2019-03-05
Remove blocks from the memory

### LINKED ISSUE #545: Remove blocks from the memory

rtsisyk:
Don't store blockchain in the memory. Keep only indexes.

Blocked by #579
Closes #545

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -47,12 +47,10 @@ use stegos_crypto::hash::*;
 use stegos_crypto::pbc::secure;
 use tokio_timer::clock;
 
-type BlockId = usize;
-
 /// A help to find UTXO in this blockchain.
 struct OutputKey {
     /// The short block identifier.
-    pub block_id: BlockId,
+    pub block_id: u64,
     /// Merkle Tree path inside block.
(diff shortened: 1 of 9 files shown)


rtsisyk:
Rebased.

blockchain/src/blockchain.rs
```diff
@@ -213,32 +212,38 @@ impl Blockchain {
...
+    pub fn contains_output(&self, output_hash: &Hash) -> bool {
+        if let Some(OutputKey { .. }) = self.output_by_hash.get(output_hash) {
+            return true;
+        }
+        return false;
+    }
+
     /// Find UTXO by its hash.
-    pub fn output_by_hash(&self, output_hash: &Hash) -> Option<&Output> {
+    pub fn output_by_hash(&self, output_hash: &Hash) -> Result<Option<Output>, Error> {
```
vldm: It would be good to remove this result, as i see in code there two possible handlers:
1) panic in case of None, or Err.
2) Throw error in case of None, or Err.

There no differences in business logic, is it io error, or Output really was not found.

If it would easy to map rocksdb::Error into BlockchainError, it would be great, if no, we can just create ticket and make it later.

Issue #576, opened 2019-03-04
Export basic metrics to Prometheus


rtsisyk:
- [ ] Epoch
- [ ] Height
- [ ] The number of UTXO (`self.chain.outputs_by_id.len()`)
- [ ] Mempool size
- [ ] Is Leader (a flag)
- [ ] Consensus State
- [ ] VRF Tickets State

PR #583, opened 2019-03-06
Add more metrics.

vldm:
Closes: #576
Closes: https://github.com/stegos/stegos/issues/575


consensus/src/state.rs
```diff
@@ -142,6 +142,7 @@ impl<Request: Hashable + Clone + Debug, Proof: Hashable + Clone + Debug> Consens
     pub fn reset(&mut self, height: u64) {
         self.height = height;
         self.state = ConsensusState::Propose;
+        crate::metrics::CONSENSUS_STATE.set(crate::metrics::ConsensusState::Propose as i64);
```
echupriyanov: Why not 'use crate::metrics`?

PR #589, opened 2019-03-07
no longer includes transaction signature in hash

dbmcclain:
No longer includes transaction signature in hash of transaction records

### DIFF
--- a/blockchain/src/transaction.rs
+++ b/blockchain/src/transaction.rs
@@ -97,6 +97,12 @@ pub struct Transaction {
     pub sig: SchnorrSig,
 }
 
+impl Hashable for Transaction {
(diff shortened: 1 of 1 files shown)


network/src/libp2p_network/mod.rs
```diff
@@ -354,6 +354,7 @@ where
                     }
                     Err(e) => error!("Failure decoding unicast message: {}", e),
                 }
+                return;
```
vldm: you probably wrongly rebased your pullrequest, because this changes was already merged in https://github.com/stegos/stegos/pull/588/files

vldm: And you have only one commit that is authored by eugene.

Issue #564, opened 2019-02-28
Test is chain consisntent


vldm:
For each of our chains, our code should work.

We need a test that will:
1) Run node on each chain and checks if it correctly started (inited genesis)
2) Run 3 nodes on dev chain, and check if they synchronized.

rtsisyk:
do you mean integration testing?

PR #606, opened 2019-03-12
Added chain consistency check.

vldm:
Closes: #564 
Test if chain can be loaded from stored genesis.


vldm:
https://github.com/stegos/stegos/pull/614

node/src/test/mod.rs
```diff
@@ -20,22 +20,35 @@
 // SOFTWARE.
 
 pub mod loopback;
+
+#[cfg(test)]
```
rtsisyk: WTF?

Issue #608, opened 2019-03-12
Add vrf to KeyBlock


vldm:
First phase of changing our consensus - is to add random to KeyBlock, in order to make "chain of random".

PR #610, opened 2019-03-12
Add random and view_change to keyblock.

vldm:
This PR adds random and view_change to KeyBlock.

Closes: #608 
Closes: #612


crypto/src/utils/mod.rs
```diff
@@ -23,12 +23,20 @@
...
 
+use crate::hash::{Hash, Hashable, Hasher};
 use hex;
 use std::cmp::Ordering;
-
 // -------------------------------------------------------------------
 // general utility functions
 
+/// Mix seed hash with round value to produce new hash.
+pub fn mix(random: Hash, round: u32) -> Hash {
```
rtsisyk: Too generic. I suggest to revert this function to original place.

blockchain/src/blockchain.rs
```diff
@@ -85,6 +86,10 @@ pub struct Blockchain {
     pub last_block_timestamp: Instant,
     /// The hash of the last block.
     pub last_block_hash: Hash,
+
+    //TODO: Remove leader,facilitator,validators, and use save only last keyblock
```
rtsisyk: No need to save entire raw key block. Let's remove this comment.

PR #613, opened 2019-03-12
produce unencrypted UTXO 

dbmcclain:
unencrypted UTXO with 

new_uncloaked(recipient_pkey: &PublicKey, amount: i64) -> Result<(Self, Fr)
  -- returns UTXO and gamma from Bulletproof

with_uncloaked_payload(
        recipient_pkey: &PublicKey,
        amount: i64,
        data: PaymentPayloadData,
    ) -> Result<(Self, Fr), Error>  -- returns UTXO and gamma from Bulletproof

Recipient pub key remains uncloaked, payload remains unencrypted, with keying hint of all zero


rtsisyk:
This implementation is very smart and doesn't have breaking changes.

PR #615, opened 2019-03-13
Migration to rust-libp2p v0.5

### LINKED ISSUE #573: Migrate to libp2p v0.4
Catch-up with `rust-libp2p`

echupriyanov:
Migrated to rust-libp2p v0.5

* replaced libp2p/floodsub with forked version (stegos/pubsub):
  * message size no longer limited
  * unsubscription is not supported
  * allow pubsub operate on all connected nodes (original keeps its own network view and works only  with explicitly allowed peers)
* number of NCP improvements, making it less aggressive in resource usage
* for now removed all DHT stuff

Closes #573


network/src/libp2p_network/mod.rs
```diff
@@ -191,52 +192,52 @@ fn new_service(
...
-        kad.add_providing(network_pkey_to_peer_id(&keychain.network_pkey));
+        // let mut discovery = DiscoveryBehaviour::new(peer_id.clone());
+        // discovery.add_providing(network_pkey_to_peer_id(&keychain.network_pkey));
         let mut behaviour = Libp2pBehaviour {
-            floodsub: floodsub::Floodsub::new(peer_id.clone()),
-            ncp: ncp::layer::Ncp::new(config),
-            kad,
+            floodsub: Floodsub::new(peer_id.clone()),
+            ncp: Ncp::new(config),
+            // unicast_send: UnicastSend::new(keychain),
```
vldm: Let keep production code without comments

PR #616, opened 2019-03-13
Auto-commit proposed block with 2/3 signatures on timeout

### LINKED ISSUE #539: Autocommit block with 2/3 precommits on timeout

rtsisyk:
Commit proposed block with 2/3 pre-commits automatically if leader hasn't done it on time.

+ Add basic tests for consensus.

Closes #539

I don't like the part of code related to multisignature. This patch tries to augment a multisignature returned by consensus with leader's signature from proposal.


rtsisyk:
Added a metric.

PR #620, opened 2019-03-13
ValueShuffle Hardening

dbmcclain:
cleanup, light refactoring, 
initial error traps and race conditions examined, 
first cut at uniform state handling on error returns

remove warning ignores and cleanup to standard Rust
remove unused code

### DIFF
--- a/wallet/protos/valueshuffle.proto
+++ b/wallet/protos/valueshuffle.proto
@@ -4,10 +4,6 @@ package stegos.valueshuffle;
 import "crypto.proto";
 import "blockchain.proto";
 
-message Example {
-    string payload = 1;
-}
-
 message DcRow {
     repeated stegos.crypto.Fr cols = 1;
 }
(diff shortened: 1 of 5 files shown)


wallet/src/lib.rs
```diff
@@ -21,7 +21,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
-#![deny(warnings)]
+// #![deny(warnings)]
```
rtsisyk: Please fix errors.

dbmcclain: what errors?

rtsisyk: Please revert `#![deny(warnings)]` and fix all warning instead.

dbmcclain: deny(warnings) reverted. Sorry, I was under the impression that this softened the Rust scrutiny and that leaving it out makes Rust more critical of small semantic errors. I guess I had it backwards? At any rate, reverted as per instructions and no errors, no warnings.

PR #637, opened 2019-03-18
Implement multi-versioned map

rtsisyk:
Needed for #629


dbmcclain:
I just did a cargo fmt and re-pushed the PR, to get past CI build errors

rtsisyk:
What do you think about this design?

dbmcclain:
I find you design *very* interesting. I wouldn't have thought of using an undo list to the side. Not sure what I might have done. Maybe I would have converged on your idea, but maybe not... Kudos!

PR #644, opened 2019-03-20
Implement the rollback operation for monetary blocks

### LINKED ISSUE #629: Add support for forking in the blockchain storage

rtsisyk:
Add a new method to `struct Blockchain` called pop_monetary_block().
This method can be use to roll back all changes caused by previous
call of push_monetary_block(). Only monetary blocks can be reverted.
Every key block finalizes the state of blockchain database and cleans
the undo log for previous monetary blocks.

+ Add basic tests cases for persistency.

Closes #629

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -30,12 +30,12 @@ use crate::error::*;
 use crate::escrow::*;
 use crate::merkle::*;
 use crate::metrics;
+use crate::mvcc::MultiVersionedMap;
 use crate::output::*;
 use crate::storage::ListDb;
 use failure::ensure;
 use failure::Error;
 use log::*;
-use std::collections::HashMap;
 use std::collections::HashSet;
 use std::time::Instant;
 use stegos_crypto::bulletproofs::fee_a;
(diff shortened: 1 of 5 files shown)


blockchain/src/blockchain.rs
```diff
@@ -994,6 +1050,63 @@ impl Blockchain {
...
+        let version = block_id;
+
+        //
+        // Remove from the disk.
+        //
+        let block = self.block_by_id(block_id)?;
+        let block = if let Block::MonetaryBlock(block) = block {
+            block
+        } else {
+            panic!("Expected monetary block");
```
vldm: This method is public, maybe return error?

blockchain/src/blockchain.rs
```diff
@@ -50,14 +50,34 @@ use tokio_timer::clock;
...
+    burned: ECp,
+    /// The total sum of gamma adjustments.
+    gamma: Fr,
+    /// The total sum of monetary adjustments.
+    monetary_adjustment: i64,
+}
+
+type BlockByHashMap = MultiVersionedMap<Hash, u64, u64>;
+type OutputByHashMap = MultiVersionedMap<Hash, OutputKey, u64>;
+type BalanceMap = MultiVersionedMap<u8, Balance, u64>;
```
vldm: Convert to `Map<(), Balance, u64>`

blockchain/src/blockchain.rs
rtsisyk: This problem can only be caused by corrupted database or a bug.

blockchain/src/blockchain.rs
rtsisyk: I'll try.

blockchain/src/blockchain.rs
vldm: Yes, this panic could be emited, only on some bug, because of this assert
`assert_ne!(block_id, self.last_key_block_id);`.  But still, this assertion could panic, if user will call `pop_monetary_block`, since this method is public interface of blockchain, let user define how to handle situation when there no monetary blocks.

PR #652, opened 2019-03-21
Remove BLS signature from Monetary Blocks

### LINKED ISSUE #651: Remove multisignature from monetary blocks

rtsisyk:
- Keep single BLS signature of leader for Monetary Blocks.
- Move signatures to block body.

Closes #651


blockchain/src/blockchain.rs
```diff
@@ -696,17 +696,8 @@ impl Blockchain {
 
         //TODO: remove multisig?
```
vldm: remove comment

rtsisyk: OK, I'm removing multisig by the next patch.

PR #653, opened 2019-03-21
Fix devnet bugs.

vldm:


node/src/loader.rs
```diff
@@ -145,19 +145,21 @@ impl NodeService {
...
-        return master;
+        return Some(master);
     }
 
     /// Request the block history from an random node.
     pub fn request_history(&mut self) -> Result<(), Error> {
-        let master = self.choose_master();
+        let master = self
+            .choose_master()
+            .ok_or_else(|| format_err!("Failed to get validator list."))?;
```
rtsisyk: This line looks buggy. I suppose that format_err!() creates a temporary error object which is immediately discarded.

PR #655, opened 2019-03-21
Remove almost all hard-coded constants

### LINKED ISSUE #646: Make TX_WAIT_TIMEOUT and SEALED_BLOCK_IN_EPOCH configurable

rtsisyk:
Move hard-coded constants to configuration files.

Closes #646

### DIFF
--- a/node/src/config.rs
+++ b/node/src/config.rs
@@ -1,0 +1,82 @@
+//
+// MIT License
+//
+// Copyright (c) 2019 Stegos AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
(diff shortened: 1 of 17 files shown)


node/src/config.rs
```diff
@@ -0,0 +1,82 @@
...
+// SOFTWARE.
+
+use serde_derive::{Deserialize, Serialize};
+
+/// Chain configuration.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(default)]
+pub struct ChainConfig {
+    /// Time delta in which our messages should be delivered, or forgotten.
+    pub message_timeout: u64,
```
vldm: Can we use Duration, for message_timeout type?

node/src/config.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Estimated time of block validation
+    pub block_validation_timeout: u64,
+    /// How long wait for transactions before starting to create a new block.
+    pub tx_wait_timeout: u64,
+    /// How long wait for blocks.
+    pub block_timeout: u64,
```
vldm: May be rename it to key_block_timeout, since for now monetary blocks, didn't need so long time.

And probably, we should skipp deserialization, and always use formula
```
let block_timeout = tx_wait_timeout + // propose timeout
            message_timeout * 3 + // 3 consensus message
            block_validation_timeout * 3; // leader + validators + sealed block.
```

node/src/config.rs
rtsisyk: I don't know how to Serialize it.

node/src/config.rs
rtsisyk: OK

Issue #664, opened 2019-03-25
Re-request blocks from the current leader on timeout


rtsisyk:
Add a temporary workaround for missing packets until view changes are implemented.
See https://github.com/stegos/stegos/issues/654

PR #669, opened 2019-03-25
Request block from last leader on timeout.

vldm:
Closes: #664

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -194,6 +194,7 @@ pub enum NodeMessage {
     Init,
     NetworkReady,
     ConsensusTimer(Instant),
+    ViewChangeTimer(Instant),
 }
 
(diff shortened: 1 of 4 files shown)


node/src/lib.rs
```diff
@@ -802,7 +810,18 @@ impl NodeService {
 
         Ok(())
     }
+    /// Request block history from leader, if no block was received
+    /// retry each message_timeout.
+    fn handle_view_change_timer(&mut self) -> Result<(), Error> {
+        let elapsed: Duration = clock::now().duration_since(self.chain.last_block_timestamp());
 
+        if self.consensus.is_none() && elapsed >= Duration::from_secs(self.cfg.message_timeout) {
```
rtsisyk: self.cfg.block_timeout instead of self.cfg.message_timeout

PR #679, opened 2019-03-28
Added encryption to unicast send

### LINKED ISSUE #671: Unicast Messages Encryption

echupriyanov:
* now we both sign outgoing messages with sender key and encrypt with
recipient key
* helper functions to encode/decode pbc::secure::RVal values

Closes #671

### DIFF
--- a/network/src/libp2p_network/mod.rs
+++ b/network/src/libp2p_network/mod.rs
@@ -60,6 +60,7 @@ pub struct Libp2pNetwork {
 }
 
 const UNICAST_TOPIC: &'static str = "stegos-unicast";
+const IBE_ID: &'static [u8] = &[105u8, 13, 185, 148, 68, 76, 69, 155];
 
 impl Libp2pNetwork {
     pub fn new(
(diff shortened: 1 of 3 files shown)


rtsisyk:
@dbmcclain please review

network/src/libp2p_network/mod.rs
```diff
@@ -546,79 +567,118 @@ fn my_external_address(config: &NetworkConfig) -> Vec<Multiaddr> {
...
-    if secure::check_hash(&hash, &sig, &from) {
-        Ok((from, to, protocol_id, data))
+fn decrypt_message(
+    my_skey: &secure::SecretKey,
+    mut payload: UnicastPayload,
+    signature: secure::Signature,
+    rval: secure::RVal,
+) -> Result<UnicastPayload, Error> {
+    let enc_packet = secure::EncryptedPacket::new(&payload.to, IBE_ID, &rval, &payload.data);
+
```
dbmcclain: okay... it looks like you are signing the plaintext of the message. But for authentication, purposes, it is generally the advice of other experts that you should sign what you sent. And since you sent an encrypted payload, you should sign that encrypted payload, along with the other plain-view fields (from, to, protocol, etc).

The issue is that sign-then-encrypt forces you to accept unauthenticated data prior to decryption, and then spend the time decrypting a possibly garbage message from some troll. By using encrypt-then-sign, you can immediately reject the message if the signature fails, even before wasting time decrypting the message.

dbmcclain: https://www.iacr.org/archive/crypto2001/21390309.pdf

dbmcclain: according to Krawczyk, sign-then-encrypt also opens you to chosen plaintext attacks.

echupriyanov: Great thanks for advice! Will update correspondingly

PR #680, opened 2019-03-28
Add WebSocket API Server

rtsisyk:
- Add a WebSocket server based on Tokio.

See #69

### DIFF
--- a/testing/node01/stegos.toml
+++ b/testing/node01/stegos.toml
@@ -10,7 +10,6 @@ private_key = "testing/node01/stegos.skey"
 public_key = "testing/node01/stegos.pkey"
 
 [network]
-node_id = "node01"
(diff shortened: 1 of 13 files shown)


testing/node01/stegos.toml
```diff
@@ -20,4 +19,8 @@ seed_nodes = [
 ]
 advertised_addresses = [
     "/ip4/127.0.0.1/tcp/10055",
-]
\ No newline at end of file
+]
+
+[wsapi]
```
vldm: It could be hard for end-user to deduct what this field mean.

Maybe simplify it to just `[api]` ?

rtsisyk: I thought about it. Let's change to API.

rtsisyk:
Re-based.

PR #681, opened 2019-03-28
Add BLS signature on PBC Public Key in Stake UTXO

dbmcclain:
Small change, but percolates upward through entire system: Blockchain, Wallet, all test code therein.

BLS Signature on the published PBC Public Validator Key shows proof that we know its corresponding secret key, and hence we have ownership of that key, regardless of the cloaked recipient key provided in the UTXO.


dbmcclain:
found that I needed tool "j2" to regenerate genesis properly.

dbmcclain:
@rtsisyk @vldm - this small change affects your code for Stake UTXO, both staking and unstacking. Just so you are aware.

PR #691, opened 2019-04-01
Refactor multisignature, to use index based.

vldm:
First iteration of refactor, to use validator id instead of public key.

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -33,9 +33,9 @@ use crate::metrics;
 use crate::mvcc::MultiVersionedMap;
 use crate::output::*;
 use crate::storage::ListDb;
-use failure::Error;
+use failure::{ensure, Error};
 use log::*;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::time::Instant;
 use stegos_crypto::bulletproofs::fee_a;
 use stegos_crypto::bulletproofs::validate_range_proof;
@@ -470,6 +470,12 @@ impl Blockchain {
         self.view_change
     }
 
+    /// Returns number of total slots in current epoch.
+    /// Internally always return cfg.max_slot_count
+    pub fn total_slots(&self) -> i64 {
+        self.cfg.max_slot_count
+    }
+
     //----------------------------------------------------------------------------------------------
     // Key Blocks
     //----------------------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


blockchain/src/blockchain.rs
```diff
@@ -471,6 +471,12 @@ impl Blockchain {
         self.view_change
     }
 
+    /// Returns number of total slots in current epoch.
+    /// Internally always return cfg.max_slot_count
+    pub fn total_slots(&self) -> i64 {
+        self.cfg.max_slot_count as i64
```
rtsisyk: Why cast is needed here? Please either change max_slot_count to i64 or use u64 everywhere.

blockchain/src/error.rs
```diff
@@ -55,6 +55,8 @@ pub enum BlockchainError {
     MissingValidators,
     #[fail(display = "The leader must be validator.")]
     LeaderIsNotValidator,
+    #[fail(display = "No leader signature was found in BLS signature.")]
+    NoLeaderSignatureFound,
```
rtsisyk: I suggest to change `check_multi_signature()` to return `Result<(), Error>`

blockchain/src/multisignature.rs
```diff
@@ -43,92 +45,98 @@ pub fn create_multi_signature(
...
-        count += 1;
-    }
+    for bit in multisigmap.iter() {
+        let validator = &validators[bit];
+        let pkey: secure::G2 = validator.0.into();
+        let slots = validator.1;
+        assert!(slots > 0);
 
-    // Multi-signature must contain leader's key.
-    if !has_leader {
```
rtsisyk: I think you can pass leader_id to this function and remove this part from validate_key_block():
```

            if let Some(leader_id) = validators_map.get(&leader) {
                // bit of leader should be always set.
                ensure!(
                    block.body.multisigmap.contains(*leader_id as usize),
                    BlockchainError::NoLeaderSignatureFound
                );
            } else {
                return Err(BlockchainError::LeaderIsNotValidator.into());
            }
```

blockchain/src/blockchain.rs
```diff
@@ -563,24 +569,35 @@ impl Blockchain {
...
-            if !check_multi_signature(
-                &block_hash,
-                &block.body.multisig,
-                &block.body.multisigmap,
-                &validators,
-                &leader,
-                is_proposal,
-            ) {
+            // ignore multisignature if it proposal.
+            if !is_proposal
```
rtsisyk: Please check proposal for the leader's signature.

blockchain/src/multisignature.rs
vldm: I think checking leader is out of responsible for multisignature.

PR #693, opened 2019-04-01
Add block_lag and block_idle metric

rtsisyk:
- Use millisecond precision in block.header.timestamp.
 - Update default timeouts.
- Use `std::time::SystemTime` instead of u64.
- Use `std::time::Duration` instead of u64.
- Remove `chrono` from dependencies.
- Unify `current_timestamp`/`timestamp` usage.
- Add block_lag and block_idle metrics.

See #688

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2272,7 +2272,6 @@ version = "0.2.0"
 dependencies = [
  "bitvector 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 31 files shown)


rtsisyk:
```
# HELP stegos_block_idle_ms The elapsed time since the last block, i.e. it is the time difference between the local time at this node and the time when the last block was registered.
# TYPE stegos_block_idle_ms gauge
stegos_block_idle_ms 3
# HELP stegos_block_lag_ms The last block creation + validation + propagation time, i.e. it is the time difference between the local time at a remote leader when the last block began to be created and the local time at this node when this block was registered.
# TYPE stegos_block_lag_ms gauge
stegos_block_lag_ms 2453
# HELP stegos_block_local_timestamp_ms The local time at this node when the last block was registered.
# TYPE stegos_block_local_timestamp_ms gauge
stegos_block_local_timestamp_ms 1554130356649
# HELP stegos_block_remote_timestamp_ms The local time at a remote leader when the last block began to be created, i.e it equals to the value of block.header.timestamp.
# TYPE stegos_block_remote_timestamp_ms gauge
stegos_block_remote_timestamp_ms 1554130354196
```

PR #702, opened 2019-04-03
Fix BulletProof quadrant redundancy

dbmcclain:
Saves 9 Fr values in BP (288 bytes),
costs 15 Hash + 12 PtDecompress in Validation time


src/stegos.rs
```diff
@@ -375,6 +375,7 @@ mod tests {
     }
 
     #[test]
+    // #[ignore] // TODO: remove this before pushing repo!
```
rtsisyk: @echupriyanov will re-generate it.

dbmcclain: I had to comment out that test again, since CI pipeline fails the test too. Re-pushed the branch

PR #739, opened 2019-04-09
Speed up network synchronization procedure

rtsisyk:
Follow up #687

### DIFF
--- a/node/src/loader.rs
+++ b/node/src/loader.rs
@@ -179,13 +179,18 @@ impl NodeService {
             response.height,
         );
 
+        let initial_height = self.chain.height();
         for block in response.blocks {
             // Fail on the first error.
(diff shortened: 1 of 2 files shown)


node/src/loader.rs
```diff
@@ -179,13 +179,18 @@ impl NodeService {
...
         }
 
-        // Request more blocks.
-        if !self.is_synchronized() {
+        //
+        // Request more blocks in the follwing cases:
+        // a) The timestamp of the latest keyblock is oudated (see is_synchronized()).
+        // b) At least two blocks have been applied.
+        //
+        if !self.is_synchronized() || (self.chain.height() >= initial_height + 2) {
```
vldm: I didn't like magic constants, can you give this constant name?

PR #744, opened 2019-04-09
Split timers from node.

vldm:
Move timers to separete structure, this will alow better control on timers in future.


node/src/lib.rs
```diff
@@ -29,12 +29,14 @@ pub mod metrics;
 pub mod protos;
 #[cfg(test)]
 mod test;
+#[macro_use]
```
rtsisyk: #[macro_use] is deprecated, replace with `use mod::macros_name`.

PR #746, opened 2019-04-09
Export Node API to WebSocket

### LINKED ISSUE #743: Export Node API to WebSocket

rtsisyk:
Closes #743

### DIFF
--- a/blockchain/src/election.rs
+++ b/blockchain/src/election.rs
@@ -37,9 +37,9 @@ pub struct ElectionInfo {
     pub height: u64,
     pub view_change: u32,
     pub slots_count: i64,
-    pub last_leader: String,
-    pub current_leader: String,
-    pub next_leader: String,
+    pub last_leader: Option<secure::PublicKey>,
+    pub current_leader: secure::PublicKey,
+    pub next_leader: secure::PublicKey,
 }
 
 /// Result of election.
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -41,11 +41,14 @@ use crate::validation::*;
 use bitvector::BitVector;
 use failure::Error;
 use futures::sync::mpsc::{unbounded, UnboundedReceiver, UnboundedSender};
+use futures::sync::oneshot;
 use futures::{Async, Future, Poll, Stream};
 use futures_stream_select_all_send::select_all;
 use log::*;
 use protobuf;
 use protobuf::Message;
+use serde_derive::Deserialize;
+use serde_derive::Serialize;
 use std::collections::BTreeMap;
 use std::time::SystemTime;
 use std::time::{Duration, Instant};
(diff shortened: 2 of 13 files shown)


blockchain/src/election.rs
```diff
@@ -37,9 +37,9 @@ pub struct ElectionInfo {
     pub height: u64,
     pub view_change: u32,
     pub slots_count: i64,
-    pub last_leader: String,
-    pub current_leader: String,
-    pub next_leader: String,
+    pub last_leader: Option<secure::PublicKey>,
```
vldm: For now our console api will return full PublicKey, is it ok?

node/src/lib.rs
```diff
@@ -83,18 +86,12 @@ impl Node {
...
-        Ok(())
-    }
-
-    /// Ask current state of the election.
-    pub fn escrow_info(&self) -> Result<(), Error> {
-        let msg = NodeMessage::EscrowInfo;
-        self.outbox.unbounded_send(msg)?;
-        Ok(())
+    /// Execute a Node Request.
+    pub fn request(&self, request: NodeRequest) -> oneshot::Receiver<NodeResponse> {
```
vldm: I better like separate `election_info`/`escrow_info` methods as api, rather than just single method with enum, can we just return tx in this methods?

node/src/lib.rs
```diff
@@ -1106,9 +1085,18 @@ impl Future for NodeService {
...
+                        NodeMessage::Request { request, tx } => {
+                            let response = match request {
+                                NodeRequest::ElectionInfo {} => {
+                                    NodeResponse::ElectionInfo(self.chain.election_info())
+                                }
+                                NodeRequest::EscrowInfo {} => {
+                                    NodeResponse::EscrowInfo(self.chain.escrow().info())
+                                }
+                            };
+                            tx.send(response).ok(); // ignore errors.
```
vldm: why not write error in log?

blockchain/src/election.rs
rtsisyk: Yes, this patch introduces a regressions to "show election" and "show escrow". I didn't want to implement a separate infrastructure for CLI. Let's keep long keys?

node/src/lib.rs
rtsisyk: I can restore re-implement these methods as wrappers on the top of request(), but they will be unused.

node/src/lib.rs

src/console.rs
```diff
@@ -532,25 +533,25 @@ impl Future for ConsoleService {
             }
         }
 
-        let wallet_responses = std::mem::replace(&mut self.wallet_responses, Vec::new());
-        for mut rx in wallet_responses {
+        if self.wallet_response.is_some() {
```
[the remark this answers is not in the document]
rtsisyk:
```
536 |         if let Some(rx) = self.wallet_response {
    |                     --    ^^^^^^^^^^^^^^^^^^^^
    |                     |     |
    |                     |     cannot move out of borrowed content
    |                     |     help: consider borrowing here: `&self.wallet_response`
    |                     data moved here
```

node/src/lib.rs
rtsisyk: This situation can only happen if remote peer has disconnected. In this case we already print "Disconnected" message to the log. I don't think that it makes sense to print warn/error message for each `oneshot` channel.

blockchain/src/election.rs
vldm: I am fine with this.

node/src/lib.rs
vldm: Why it should be unused?


You can use it in console 

```
            let request = NodeRequest::ElectionInfo {};
            self.node_response = Some(self.node.request(request));
```
And probably in websockets too?

src/console.rs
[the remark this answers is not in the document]
vldm: ` if let Some(ref rx) = self.wallet_response.take()`

or
` if let Some(rx) = &self.wallet_response `
should work

node/src/lib.rs
vldm: @rtsisyk sorry, i missed can you refer to line, where we printing "Disconnected" message at exit. I only found `panic!("disconnect")` on other side of the channel.
For me it's better to know when api job is disconnected, intstead of debug why some message are not being processed. You can write it as `info` or even `debug` if you suppose this event is frequently.

rtsisyk: It makes absolutely no sense.

node/src/lib.rs
rtsisyk: I don't want to support two interfaces for the same thing.

PR #748, opened 2019-04-10
Fix error in decrypt_key(), and add transient pubkey to result of enc…

dbmcclain:
…ryption for use by public validation utillities

### DIFF
--- a/blockchain/src/output.rs
+++ b/blockchain/src/output.rs
@@ -33,7 +33,6 @@ use stegos_crypto::curve1174::fields::Fr;
 use stegos_crypto::curve1174::G;
 use stegos_crypto::hash::{Hash, Hashable, Hasher, HASH_SIZE};
 use stegos_crypto::pbc::secure;
-use stegos_crypto::CryptoError;
 
 /// A magic value used to encode/decode payload.
 const PAYMENT_PAYLOAD_MAGIC: [u8; 4] = [112, 97, 121, 109]; // "paym"
@@ -138,7 +137,7 @@ fn cloak_key(
     recipient_pkey: &PublicKey,
     gamma: &Fr,
     timestamp: SystemTime,
-) -> Result<(PublicKey, Fr), CryptoError> {
+) -> Result<(PublicKey, Fr), Error> {
     // h is the digest of the recipients actual public key mixed with a timestamp.
     let mut hasher = Hasher::new();
     recipient_pkey.hash(&mut hasher);
--- a/blockchain/src/transaction.rs
+++ b/blockchain/src/transaction.rs
@@ -313,9 +313,9 @@ impl Transaction {
         let eff_pkey: PublicKey = eff_pkey.into();
 
         // Check signature
-        match validate_sig(&tx_hash, &self.sig, &eff_pkey)? {
-            true => Ok(()),
-            false => Err(TransactionError::InvalidSignature(tx_hash).into()),
+        match validate_sig(&tx_hash, &self.sig, &eff_pkey) {
+            Ok(_) => Ok(()),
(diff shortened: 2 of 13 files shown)


crypto/src/curve1174/cpt.rs
```diff
@@ -580,11 +580,16 @@ pub fn decrypt_key(
...
     // then decrypt the key
+    //
+    // Return CryptoError::BadKeyingSignature if the Schnorr
+    // signature fails to validate the hash of the encrypted payload.
     let (skey, pkey) = make_securing_keys(seed);
     let mut state = Hasher::new();
     encr_key.hash(&mut state);
     let h = state.result();
-    validate_sig(&h, chk, &pkey)?;
+    if !validate_sig(&h, chk, &pkey)? {
```
vldm: `validate_sig` has realy not user friendly API, it returns Error, in case of decompress failure, and returns bool, for validation result. It is very easy to forgot second check.

Can we modify `validate_sig` method, to return `Result<(), Error>`, and hide `CryptoError::BadKeyingSignature` inside this method?

dbmcclain: Yes, I agree with you. I have changed validate_sig() to return either Ok(true) or Err(CryptoError::BadKeyingSignature)

dbmcclain: ... and let's stop tripping over our own shoelaces... I converted entire crypto lib to use failure::Error instead of specific CryptoError.

blockchain/src/output.rs
```diff
@@ -138,7 +137,7 @@ fn cloak_key(
     recipient_pkey: &PublicKey,
     gamma: &Fr,
     timestamp: SystemTime,
-) -> Result<(PublicKey, Fr), CryptoError> {
+) -> Result<(PublicKey, Fr), Error> {
```
rtsisyk: Replacing `CryptoError` with `Error` is not necessary. Generic `Error` object is only needed if a function can return different type of error objects (e.g. NodeError and CryptoError)

blockchain/src/transaction.rs
```diff
@@ -313,9 +313,9 @@ impl Transaction {
         let eff_pkey: PublicKey = eff_pkey.into();
 
         // Check signature
-        match validate_sig(&tx_hash, &self.sig, &eff_pkey)? {
-            true => Ok(()),
-            false => Err(TransactionError::InvalidSignature(tx_hash).into()),
+        match validate_sig(&tx_hash, &self.sig, &eff_pkey) {
```
rtsisyk: You can try `validate_sig(&tx_hash, &self.sig, &eff_pkey).map_err(|_e| TransactionError::InvalidSignature(tx_hash))` here

wallet/src/valueshuffle/mod.rs
```diff
@@ -1656,8 +1657,8 @@ impl ValueShuffle {
 
         // check signature on this portion of transaction
         match validate_sig(&hash, &sig, &eff_pkey) {
```
rtsisyk: Try .is_ok()

rtsisyk:
I don't see the fix itself. Please push a patch with fix.

crypto/src/curve1174/cpt.rs
```diff
@@ -459,11 +459,15 @@ pub fn sign_hash_with_kval(
     }
 }
 
-pub fn validate_sig(hmsg: &Hash, sig: &SchnorrSig, pkey: &PublicKey) -> Result<bool, CryptoError> {
+pub fn validate_sig(hmsg: &Hash, sig: &SchnorrSig, pkey: &PublicKey) -> Result<(), CryptoError> {
```
dbmcclain: This is the code fix for the error in signature checking

PR #754, opened 2019-04-15
Add generator tx implementation

vldm:
This pullrequest add transaction generator possibility to stegos node.

To start generator one can simply adds line
```
generate_txs = ["84a7e8b14b66b4567fce12b5589a93f2f6ed4404e86f3adb148ca981fb6c7a54"]
```
to `[general]` config section.

This line will activate generator to `84a7e8b14b66b4567fce12b5589a93f2f6ed4404e86f3adb148ca981fb6c7a54`

Generator will choose random key from `generate_txs` field, and create transaction for it.
One transaction at a time.

For cli usage, one can use `generator` command, with `stop` and `start` subcommands.

To add more key use `generator start NEW_KEY,NEW_KEY2`

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -1,3 +1,4 @@
+use crate::config::GeneralConfig;
 ///! Console - command-line interface.
 //
 // Copyright (c) 2018 Stegos AG
@@ -20,6 +21,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 use crate::consts;
+use crate::generator::Generator;
 use dirs;
 use failure::Error;
 use futures::sync::mpsc::UnboundedReceiver;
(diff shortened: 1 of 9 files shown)


src/console.rs
```diff
@@ -89,18 +93,28 @@ pub struct ConsoleService {
...
         let (tx, rx) = channel::<String>(1);
         let wallet_notifications = wallet.subscribe();
         let wallet_response = None;
         let node_response = None;
         let stdin_th = thread::spawn(move || Self::readline_thread_f(tx));
         let stdin = rx;
         let unicast_rx = network.subscribe_unicast(CONSOLE_PROTOCOL_ID)?;
-
+        let generator = if !cfg.generate_txs.is_empty() {
+            Some(Generator::new(wallet.clone(), cfg.generate_txs.clone()))
```
rtsisyk: I'm not sture that ConsoleService is the right place for this generator. Try to use WalletService or a separate `Future`

vldm: Generator is seperate Future, generator is actually consumer of our wallet/node api.

For now i decide to leave near other consumer `ConsoleService` because of it's limited functional, but i am ok to split it into separate crate.

PR #762, opened 2019-04-16
Allow load more blocks.

vldm:


node/src/config.rs
```diff
@@ -38,6 +38,8 @@ pub struct ChainConfig {
     pub bonding_time: Duration,
     /// Max difference in timestamps of leader and validators.
     pub blocks_in_epoch: u64,
+    /// Loader will send maximum N epoch at time.
+    pub chain_loader_speed_in_epoch: u64,
```
rtsisyk: Looks like a temporary hack. We need to increase the length of eoch from 5 blocks to 60 or more. Anyway, this configuration option is internal and doesn't affect the user-visible behaviour, so I'm OK.

PR #782, opened 2019-04-22
Split 1 STG to 1_000_000 microSTG

rtsisyk:
This patch splits 1 STG into 1_000_000 microSTG.
Internally microSTG is used as an unit.
REPL allow to specify fractional STG value.

See #763


rtsisyk:
A breaking change, moving to testnet-6.

Issue #725, opened 2019-04-08
Implement key-store crypto support


dbmcclain:
Implement a provisional key storage and retrieval library for use by nodes, along the lines of SSH key storage

PR #786, opened 2019-04-23
Refactor keying

rtsisyk:
- Don't returni the third result (a signature) from make_secure_keys().
  This result is never used.
- Return Result<(), Error> from check_hash().
- Rework check_keying() to work without a signature.
- Add EncryptedKey structure and protobuf definitions.

Needed for #725
Needed for #784


dbmcclain:
what do you mean the signature is never used? It is used when you re-import and attempt to decrypt a stored signature.

PR #806, opened 2019-04-29
Slashing basic implementation.

vldm:
For now, work on this feature is stopped untill refactor of microblock will be merged into dev.


blockchain/protos/blockchain.proto
```diff
@@ -33,6 +33,13 @@ message Output {
     }
 }
 
+message SlashingTransaction {
+    SlashingProof proof = 1;
+    stegos.crypto.SecurePublicKey leader = 2;
```
rtsisyk: I don't see how `leader` field can be used. You can always calculate leader from `view_change` number in the block. In similar case in `CoinbaseTransaction` we don't store leader too.

node/src/validation.rs
```diff
@@ -94,7 +94,7 @@ pub(crate) fn validate_transaction(
     }
 
     // Check the monetary balance, Bulletpoofs/amounts and signature.
-    let staking_balance = tx.validate(&inputs)?;
+    let staking_balance = tx.validate(&inputs, &chain)?;
```
rtsisyk: Currently only PaymentTransaction and RestakeTransaction can be received from the network (see handle_transaction()):

https://github.com/stegos/stegos/blob/da7d3f23af86a74cb1a70788ae09abc419498e09/node/src/lib.rs#L387-L392):

I don't think that we are ready to allow SlashingTransaction via network. Let's simplify all things and inline `Transaction::validate()` into `validate_transaction()`.

Instead of calling Transaction::validate()

https://github.com/stegos/stegos/blob/da7d3f23af86a74cb1a70788ae09abc419498e09/blockchain/src/validation.rs#L357-L368

just write

```
        match &tx {
            Transaction::PaymentTransaction(_tx) => tx.validate(inputs),
            Transaction::RestakeTransaction(_tx) => tx.validate(inputs),
            _ => unreachable()!
        }; 
```

rtsisyk: I also see that the upper part of handle_transaction() should be moved to validate_transaction(), but let's do it by a separate commit before or after your patch.

blockchain/src/blockchain.rs
```diff
@@ -612,10 +621,44 @@ impl Blockchain {
         self.view_change_proof = None;
     }
 
+    ///
+    /// Check if some of validator was caught on cheating in current epoch.
+    /// Returns proof of cheating.
+    ///
+    pub fn validator_wallet(&self, peer: &pbc::PublicKey) -> Option<&PublicKey> {
+        self.escrow.validators_wallets.get(peer)
```
rtsisyk: Who checks that one Network key can be used only by one wallet? Could you please add this check to stake UTXO validation. Probably my approach with SlashingBalance will stop to work and you need a better solution. Try to implement it by a separate patch. Add information about wallet key to ValidatorInfo to test by `show escrow` command.

blockchain/src/blockchain.rs
```diff
@@ -568,6 +568,15 @@ impl Blockchain {
         self.escrow.get(validator_pkey, self.epoch)
     }
 
+    ///
+    /// Get staked value for validator.
+    ///
+    /// Returns (active_balance, expired_balance) stake.
+    ///
+    #[inline]
+    pub(crate) fn get_utxos(&self, validator_pkey: &pbc::PublicKey) -> (Vec<Hash>, i64) {
```
rtsisyk: Blockchain::get_utxos() is misleading. Please try to find a better name, preferably without `get_` prefix.

blockchain/src/blockchain.rs
```diff
@@ -966,6 +1010,19 @@ impl Blockchain {
                     gamma += tx.gamma;
                 }
                 Transaction::RestakeTransaction(_tx) => {}
+                Transaction::SlashingTransaction(tx) => {
+                    info!(
+                        "Found slashing transaction, removing validator, from list: cheater={}",
```
rtsisyk: Are you sure? It can change election results for past blocks.

blockchain/src/escrow.rs
```diff
@@ -43,12 +44,14 @@ struct EscrowValue {
...
 }
 
+type ValidatorsWallets = MultiVersionedMap<pbc::PublicKey, PublicKey, u64>;
 type EscrowMap = MultiVersionedMap<EscrowKey, EscrowValue, u64>;
 
 #[derive(Debug, Clone)]
 pub struct Escrow {
     /// Stakes.
     escrow: EscrowMap,
+    pub validators_wallets: ValidatorsWallets,
```
rtsisyk: `wallet_by_network_key` ?

blockchain/src/blockchain.rs
```diff
@@ -612,10 +621,44 @@ impl Blockchain {
...
+    ) -> Result<ElectionResult, BlockchainError> {
+        if height <= self.last_macro_block_height() {
+            return Err(BlockError::ElectionResultForPastEpoch(
+                height,
+                self.last_macro_block_height(),
+            )
+            .into());
+        }
+
+        if height >= self.height() {
```
rtsisyk: I think that `height` should be checked by upper levels and all `return Err()` here should be replaced by asserts.

blockchain/src/escrow.rs
rtsisyk: BTW, why is this map actually needed? Just take any StakeUTXO for `network_pkey` in `escrow` map and extract `recipient_pkey` from it. But remember, that you still need to check that `network_pkey` <=> `wallet_pkey` mapping is 1:1 or *:1. *:1 can be checked just be iterating `escrow: EscrowMap` by `network_pkey` (should be cheap)  and comparing `recipient_pkey` in utxos.

blockchain/src/escrow.rs
```diff
@@ -169,6 +189,33 @@ impl Escrow {
         (active_balance, expired_balance)
     }
 
+    ///
+    /// Returns list of utxos for specific validator.
+    ///
+    pub fn get_utxos(&self, validator_pkey: &pbc::PublicKey, epoch: u64) -> (Vec<Hash>, i64) {
```
rtsisyk: Try to find better naming for get() and get_utxos().

blockchain/src/escrow.rs
```diff
@@ -234,10 +281,12 @@ impl Escrow {
     #[inline]
     pub fn checkpoint(&mut self) {
         self.escrow.checkpoint();
+        self.escrow.checkpoint();
```
rtsisyk: A dup.

blockchain/src/escrow.rs
```diff
@@ -234,10 +281,12 @@ impl Escrow {
...
     }
 
     #[inline]
     pub fn rollback_to_version(&mut self, to_version: u64) {
         self.escrow.rollback_to_version(to_version);
+        self.validators_wallets.rollback_to_version(to_version);
```
rtsisyk: Above I suggested to remove self.validators_wallets to avoid data duplication.

blockchain/src/output.rs
```diff
@@ -71,7 +71,7 @@ pub enum OutputError {
 }
 
 /// Payment UTXO.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
```
rtsisyk: Let's check Hash::digest(&utxo) instead...

blockchain/src/slashing.rs
```diff
@@ -0,0 +1,157 @@
...
+        if self.block1.base.height != self.block2.base.height {
+            return Err(SlashingError::DifferentHeight(
+                self.block1.base.height,
+                self.block2.base.height,
+            )
+            .into());
+        }
+
+        if height > blockchain.height() {
+            return Err(SlashingError::InvalidProofHeight(height, blockchain.height()).into());
```
rtsisyk: I see that height is already checked here, so all error checks in election_result_by_height() can be replaced by asserts.

blockchain/src/slashing.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        if self.block1.base.view_change != self.block2.base.view_change {
+            return Err(SlashingError::DifferentLeader(
+                self.block1.base.view_change,
+                self.block2.base.view_change,
+            )
+            .into());
+        }
+
+        if Hash::digest(&self.block1) == Hash::digest(&self.block2) {
```
rtsisyk: I see that block hash is calculated two times here.

blockchain/src/slashing.rs
```diff
@@ -0,0 +1,157 @@
...
+        .iter()
+        .map(|(k, _v)| *k)
+        .filter(|k| k != cheater)
+        .collect();
+
+    if validators.is_empty() {
+        return Err(SlashingError::LastValidator(*cheater).into());
+    }
+
+    if inputs.is_empty() {
```
rtsisyk: How it can happen?

blockchain/src/transaction.rs
```diff
@@ -23,13 +23,48 @@
...
+//--------------------------------------------------------------------------------------------------
+// Slashing Transaction.
+//--------------------------------------------------------------------------------------------------
+
+/// Transaction that confiscate stake from cheater.
+#[derive(Debug, Clone)]
+pub struct SlashingTransaction {
+    pub proof: SlashingProof,
+
+    pub leader: pbc::PublicKey,
```
rtsisyk: I pointed above that `leader` is not needed.

blockchain/src/transaction.rs
```diff
@@ -537,3 +584,8 @@ impl From<RestakeTransaction> for Transaction {
         Transaction::RestakeTransaction(tx)
     }
 }
+impl From<SlashingTransaction> for Transaction {
```
rtsisyk: Missing empty line before. Again! Please check IDE configuration.

blockchain/src/validation.rs
```diff
@@ -347,14 +348,114 @@ impl RestakeTransaction {
...
+
+        // recreate transaction
+        let tx = confiscate_tx(blockchain, leader, self.proof.clone())?;
+
+        // found incorrect formed slashing transaction.
+        if tx.txins != self.txins {
+            bail!("Incorrect txins list")
+        }
+        // found unhonest devided stake.
+        for txs in tx.txouts.iter().zip(self.txouts.iter()) {
```
rtsisyk: I don't understand why this part is needed. Re-created SlashingTransaction via  `confiscate_tx` and check that it is the same. Why additional checks are needed?

BTW, confiscate_tx() should be SlashingTransaction::new().

blockchain/src/validation.rs
```diff
@@ -347,14 +348,114 @@ impl RestakeTransaction {
...
+        // Calculate the pedersen commitment difference in order to check the monetary balance:
+        //
+        //     pedersen_commitment_diff = \sum C_i - \sum C_o - fee * A
+        //
+        // Calculate `P_eff` to validate transaction's signature:
+        //
+        //     P_eff = pedersen_commitment_diff + \sum P_i
+        //
+
+        let mut txin_sum = ECp::inf();
```
rtsisyk: Are you double-checking that confiscate_tx() a.k.a SlashingTransaction::new() doesn't use the same UTXO multiple times? Please choose one approach - either re-creation + comparison or manual validation.

blockchain/src/validation.rs
```diff
@@ -347,14 +348,114 @@ impl RestakeTransaction {
...
+
 impl Transaction {
     /// Validate the monetary balance and signature of transaction.
     ///
     /// # Arguments
     ///
     /// * - `inputs` - UTXOs referred by self.body.txins, in the same order as in self.body.txins.
     ///
-    pub fn validate(&self, inputs: &[Output]) -> Result<StakingBalance, Error> {
+    pub fn validate(
```
rtsisyk: I see that this interface doesn't work. Please inline into validate_transaction() and validate_micro_block_tx() and remove this method.

node/src/lib.rs
```diff
@@ -488,7 +493,12 @@ impl NodeService {
                   self.chain.last_block_hash());
 
             metrics::CHEATS.inc();
-            // TODO: implement slashing.
+
+            let proof = SlashingProof::new_unchecked(remote.clone(), local);
+
+            if let Some(_proof) = self.cheating_proofs.insert(leader, proof) {
+                debug!("Cheater was already detected: cheater = {}", leader);
```
rtsisyk: Probably warn!() or at least info!()?

node/src/lib.rs
```diff
@@ -468,10 +473,10 @@ impl NodeService {
 
         // check multiple blocks with same view_change
         if remote_view_change == local.base.view_change {
-            if local_hash == local_hash {
```
rtsisyk: Seems to be a bug. Why Rust doesn't detect it? Let's fix it and push into dev ASAP.

node/src/mempool.rs
```diff
@@ -178,6 +179,21 @@ impl Mempool {
         let mut transactions: Vec<Transaction> = Vec::new();
         // Reserve a place for coinbase.
         transactions.push(Transaction::CoinbaseTransaction(Default::default()));
+        debug!(
+            "Processing {} slashing transactions.",
+            cheating_proofs.len()
+        );
+        for (_cheater, tx) in cheating_proofs {
```
rtsisyk: Why just don't put this transaction into mempool? I'm not sure that this idea works, but please check.

node/src/test/mod.rs
```diff
@@ -357,9 +361,10 @@ trait Api<'p> {
         let vrf = self.node(&first_leader_pk)?.create_vrf();
         let mut election = self.first_mut().node_service.chain.election_result();
         election.random = vrf;
-        Some(election.select_leader(view_change + 1))
+        Some(election.select_leader(view_change))
```
rtsisyk: Kind of magic. What is going on here?

blockchain/src/output.rs
```diff
@@ -573,6 +573,24 @@ impl Output {
     }
 }
 
+impl From<PaymentOutput> for Output {
```
rtsisyk: Yeah, I also added this code in https://github.com/stegos/stegos/pull/900/files#diff-92b1c9343dca9c5b5b9d740997415445R137 :)

blockchain/src/slashing.rs
vldm: Malicious create slashing transaction for some node that was not validator.

node/src/lib.rs
vldm: Why you need additional info, it could be just a dupe of block broadcasted.

node/src/test/mod.rs
vldm: This code was bugy, probably the bug was introduced during refactoring of view_change, but because it was never used, the bug was not found.

node/src/lib.rs
vldm: Probably clippy will detect it.

PR #808, opened 2019-04-29
Fix importation of compressed points to check for invalid subgroup, i…

dbmcclain:
…n addition to small subgroup attack, and not on curve attack.


dbmcclain:
The test for subgroup membership was particularly costly, as we have to multiply by the Fr modulus (a huge ragged integer). And then the results of the multiplication are tested and discarded. This happened on every point decompression - which is every time an ECC point (public key) was deserialized.

So rather than doing that, we now pre-scale all points by 1/4 on the way to compressed form, and re-inflate decompressed points by 4 using two doublings (no expensive multiply). This also means that, except for the point at infinity in a compressed form, and points that specifically don't lie on the curve, all other encodings are valid and will automatically decompress into the principal subgroup.

crypto/src/curve1174/cpt.rs
```diff
@@ -79,7 +79,8 @@ impl Pt {
     /// Create from an uncompressed point.
     #[inline]
     pub fn compress(ept: ECp) -> Pt {
-        let bytes = ECp::to_bytes(&ept);
+        let pt_4 = ept / Fr::from(4);
```
rtsisyk: What is it? Why it changes genesis block?

dbmcclain: The compressed points now look like they belong to ECC points that are scaled by 1/4 from what they were before. This factor gets reinflated on point decompression using fast ECC arithmetic instead of scalar multiply.

TL;DR - the encoding of all Public Keys changes, even though they represent the same effective values as before.

dbmcclain: I just provided special purpose (fast) ECC point rescaling by Fr(1/4). No more constant overhead of conversion from int64 to Fr, then Fr::invert(), then conversion to WindowVector. Just uses a constant WindowVector for that scale factor and does a fast raw ECC multiply.

dbmcclain: See my previous note about the extreme overhead in our previous point compression / decompression

dbmcclain:
eh? Can't rebase with binary formats? How to fix these conflicts?

rtsisyk:
Please rebase.

rtsisyk:
MErged

PR #825, opened 2019-05-03
Fix Pt::Ord() 

dbmcclain:
Use decompression then recompress to canonical form. Then bitwise comparison.

This method was chosen because DiceMix needs a pseudo ordering of Public Keys and Pts


dbmcclain:
https://twist.com/a/39362/ch/95451/t/829211/

PR #832, opened 2019-05-06
Thorough cleanup of base layer in Curve1174 crypto 

dbmcclain:


dbmcclain:
This code *does not* remove the Y parity bit from the compressed representation. We actually need it to overcome the reality of Sqrt(), which behaves differently from what you might expect from a theoretical perspective.

But the code represents a thorough cleaning of the base level in our Curve1174 crypto. It ensures that high level arithmetic expressions don't inadvertently mutate operands when they should not.

It also restores the PBC Signature check at startup on the validity of the Curve Parameters. The previous code merely checked that the hash of the parameter values equaled some precomputed hash checksum. But that is too weak, and allows anyone to fiddle with the curve parameters and compensate by computing a new checksum. The signature check prevents that. 

It is *exceedingly important* that the curve parameters be as they are under signature. All else in our entire blockchain depends on correctly established curve parameters for Curve1174.

rtsisyk:
Let's merge it after tagging testnet-7.

rtsisyk:
Why genesis has been changed?

PR #833, opened 2019-05-06
New blocks

rtsisyk:
Add Macro and Micro blocks.

See #828


blockchain/protos/blockchain.proto
```diff
@@ -42,10 +42,10 @@ message BaseBlockHeader {
     stegos.crypto.VRF random = 6;
 }
 
-message MicroBlockHeader {
+message MacroBlockHeader {
     BaseBlockHeader base = 1;
     stegos.crypto.Fr gamma = 2;
-    int64 monetary_adjustment = 3;
+    int64 block_reward = 3;
     stegos.crypto.Hash inputs_range_hash = 4;
```
vldm: Could block reward be negative?

blockchain/src/block.rs
```diff
@@ -94,53 +100,193 @@ impl Hashable for BaseBlockHeader {
...
+            block_fee,
+            gamma,
+            outputs,
+        };
+
+        MicroBlock::new(base, view_change_proof, coinbase, transactions, pkey)
+    }
+
+    /// Sign block using leader's signature.
+    pub fn sign(&mut self, skey: &secure::SecretKey, pkey: &secure::PublicKey) {
```
vldm: Why we need separate method for signing, cant we sign messages in constructors?

blockchain/src/blockchain.rs
```diff
@@ -62,12 +63,12 @@ impl ChainInfo {
     /// Create ChainInfo from micro block.
     /// ## Panics
     /// if view_change is equal to 0
-    pub fn from_micro_block(block: &MicroBlock) -> Self {
-        assert_ne!(block.header.base.view_change, 0);
+    pub fn from_block(base_header: &BaseBlockHeader) -> Self {
```
vldm: Chain info is used only for viewchanging in microblock, why we now need constructor from macroblock?

blockchain/protos/blockchain.proto
rtsisyk: No. There was idea that slashing should burn money. This idea is not actual anymore.

blockchain/src/blockchain.rs
rtsisyk: I'm not sure. Probably it is just refactoring bug. Anyway, this method only needs BaseBlockHeader, so requiring entire MicroBlock here is not necessary.

blockchain/src/block.rs
rtsisyk: I found that in most tests cases we don't need signed MicroBlock. Moreover, we can't sign MacroBlock in constructors. I decided to create a separate method to unify Micro and MacroBlocks. I'm not sure that it was right idea.

blockchain/src/blockchain.rs
vldm: Requiring MicroBlock here is just make API more strict.

rtsisyk: I prefer to don't require data which I don't need. This method only needs view_change, height and previous hash. Why I should force user to pass entire MicroBlock here?

rtsisyk:
Thanks!

PR #835, opened 2019-05-06
New StakeOutput UTXO format, cloaking hint in PaymentOutput

dbmcclain:
initial commit with redefined StakeOutput, modifed PaymentOutput to hold cloaking hint

removed a lot of excess code by making functions validate(), and get_pedersen_commitment() for both PaymentOutput and StakeOutput record types.

Function validate() does the basic stuff you need..
1. For PaymentOutput - check valid recipient key, valid cloaking_hint, valid Bulletproof, and valid length on encrypted payload
2. For StakeOutput - check valid recipient key, valid Validator PKey, valid amount, BLS signature validation on entire UTXO contents

Added function PaymentOutput::is_my_utxo() for checking ownership of UTXOs with given public and secret keys


dbmcclain:
Massive storage infrastructure rework to provide wipe-clean on Drop for all sensitive crypto Field items. Secret keys, everything derived from secret keys, gamma, delta, random value keying hints, etc.

dbmcclain:
We're starting to get a bit long in the tooth with this PR. Already, in rebasing against dev branch, I found so many major changes to the blockchain.rs, transaction.rs, and validation.rs, that it took more than an hour to accept the dev branch version, then re-edit to accommodate the evolution in this PR.

I think it would be good to merge into dev sooner than later, so that everyone can be working from the same sheet of music...

dbmcclain:
Entire codebase brought into pass-by-reference semantics on all large composite crypto items - which means just about everything.

rtsisyk:
Please do one thing by one commit. 

**StakeUTXO changes**: 

The new StakeUTXO format looks good.

**PaymentUTXO changes**:

What is `cloaking_hint` and how it is used?

**Removal of Copy from crypto primitives**:

As far as I understand, the initial idea was to add zapping of memory on Drop() for every crypto primitive in order to secure this system. Despite the fact that this intent was good, implementation is terrible. I expect that `a + b + c` will be written as `a + b + c`, not as `&(&a + &b) + &c`. Yes, it can be caused by limitation of Rust Language, but that only means that Rust is not ready for that. Let's keep crypto primitives as it and try to find a better solution later.  We don't have programming programming `&(&a + &b) + &c` here.

echupriyanov:
> Please do one thing by one commit.
> 
> **StakeUTXO changes**:
> 
> The new StakeUTXO format looks good.
> 
> **PaymentUTXO changes**:
> 
> What is `cloaking_hint` and how it is used?
> 
> **Removal of Copy from crypto primitives**:
> 
> As far as I understand, the initial idea was to add zapping of memory on Drop() for every crypto primitive in order to secure this system. Despite the fact that this intent was good, implementation is terrible. I expect that `a + b + c` will be written as `a + b + c`, not as `&(&a + &b) + &c`. Yes, it can be caused by limitation of Rust Language, but that only means that Rust is not ready for that. Let's keep crypto primitives as it and try to find a better solution later. We don't have programming programming `&(&a + &b) + &c` here.

Actually, as I demonstrated in Twist thread, its possible to avoid extra referencing and and parens and have code look like this: `&a + &b + &c * &d`

rtsisyk:
Let's keep Copy for now and find a better solution.

dbmcclain:
Actually, because we have Drop trait, you cannot compile `&a + &b * &c`. That requires `&a + &(&b * &c)`

echupriyanov:
> Actually, because we have Drop trait and only ref->element operators, you cannot compile `&a + &b * &c`. That requires `&a + &(&b * &c)`
> 

Its not so. Please, read carefully the Twist thread.

vldm:
Why this pullrequest contain all this refactoring?? Can we split StakeUTXO refactoring and all reference refactoring?
Because i am not sure that reference/clone refactoring is needed at all.

If you have your data on stack, LLVM can copy data internally even if your implement Drop, and didn't implement Copy. Having `move semantic` in highlevel language as rust, does not mean that data will be moved on machine level.

dbmcclain:
Unencrypted payload capabilities have been removed

dbmcclain:
Transaction renamed to PaymentTransaction. New RestakeTransaction. Provisionally plugged into your event handler with a separate topic. Dummy handler stubs.
Crypto is back to the way it was before, as far as you are concerned.

Passes all tests.

blockchain/src/blockchain.rs
```diff
@@ -815,14 +811,14 @@ impl Blockchain {
         // Global monetary balance.
         let orig_balance = self.balance();
         let balance = Balance {
-            created: orig_balance.created + created,
-            burned: orig_balance.burned + burned,
-            gamma: orig_balance.gamma + block.header.gamma,
+            created: &orig_balance.created + &created,
```
rtsisyk: Please remove & refactoring from this patch.

blockchain/src/blockchain.rs
```diff
@@ -968,16 +964,15 @@ pub fn create_fake_micro_block(
             .output_by_hash(&input_hash)
             .expect("no disk errors")
             .expect("exists");
+        input.validate().expect("valid UTXO");
         match input {
             Output::PaymentOutput(ref o) => {
                 let payload = o.decrypt_payload(&keys.wallet_skey).unwrap();
-                gamma += payload.gamma;
+                gamma += &payload.gamma;
```
rtsisyk: Please remove & refactoring from this patch.

blockchain/src/blockchain.rs
```diff
@@ -996,14 +991,12 @@ pub fn create_fake_micro_block(
         )
         .expect("keys are valid");
         outputs.push(Output::PaymentOutput(output));
-        gamma -= output_gamma;
+        gamma -= &output_gamma;
```
rtsisyk: Please remove from the patch.

blockchain/src/output.rs
```diff
@@ -559,29 +520,42 @@ impl Output {
...
     }
+
+    pub fn validate(&self) -> Result<(), Error> {
+        match self {
+            Output::PaymentOutput(o) => o.validate(),
+            Output::StakeOutput(o) => o.validate(),
+        }
+    }
+
+    pub fn get_recipient_pkey(&self) -> Result<ECp, CryptoError> {
```
rtsisyk: I suggest to remove `get_` prefix: `pub fn get_recipient_pkey(&self) -> Result<ECp, CryptoError> `

blockchain/src/output.rs
```diff
@@ -559,29 +520,42 @@ impl Output {
...
+        match self {
+            Output::PaymentOutput(o) => o.get_recipient_pkey(),
+            Output::StakeOutput(o) => o.get_recipient_pkey(),
+        }
+    }
+
+    pub fn get_pedersen_commitment(&self) -> Result<ECp, CryptoError> {
```
rtsisyk: => `pub fn pedersen_commitment(&self)`

blockchain/src/output.rs
```diff
@@ -647,8 +621,8 @@ pub mod tests {
         let (skey, pkey) = make_random_keys();
 
         // With empty comment.
-        let gamma: Fr = Fr::random();
-        let delta: Fr = Fr::random();
+        let gamma: Fr = Fr::random().make_safe();
```
rtsisyk: Please remove from this patch.

blockchain/src/transaction.rs
```diff
@@ -198,11 +310,11 @@ impl Transaction {
     ///
     pub fn new_super_transaction(
         skey: &SecretKey,
```
rtsisyk: Please remove `&` refactoring from this patch.

blockchain/src/validation.rs
```diff
@@ -560,14 +474,14 @@ impl Blockchain {
         //
         let orig_balance = self.balance();
         let balance = Balance {
-            created: orig_balance.created + created,
-            burned: orig_balance.burned + burned,
-            gamma: orig_balance.gamma + block.header.gamma,
+            created: &orig_balance.created + &created,
```
rtsisyk: Please remove `&` refactoring from this patch.

blockchain/src/blockchain.rs
```diff
@@ -707,21 +705,18 @@ impl Blockchain {
                     // Remove from the block.
                     match body.outputs.lookup(&path) {
                         Some(o) => {
+                            o.validate()?;
+                            burned += &o.get_pedersen_commitment()?;
```
rtsisyk: Please remove & refactoring from this patch.

dbmcclain:
PBC Public Key added to RestakeTransaction - so you can at least validate and know who to blame. 
Transaction restored from PaymentTransaction (it's what you guys are used to)
& adornments removed
make_safe() removed from tests only.

dbmcclain:
Remove PBC public key from RestakeTransaction
Check to be sure all TXIN carry the same PBC pubkey as the creator of the RestakeTransaction during construction

dbmcclain:
PublicPaymentOutput, a variant of Output is here. Wired in everywhere, but not yet functional in wallet.
Timestamps no longer needed for UTXO creation
Network SecretKey used instead of wallet SecretKey in UTXO creation

rtsisyk:
Extracted relevant parts, fixed bugs  and pushed as ef773bea978a86506fcd2cb818c8647497606d47:

>    - Re-define StakeOutput.
>    - Modify PaymentOutput to hold cloaking hint (not used yet).
>    - Create PublicPaymentUTXO (not used yet).
>    - Remove secret key and timestamp from UTXO construction.

- Restaking Transaction should be reworked on the top of https://github.com/stegos/stegos/pull/857.
- Usage of cloaking hint should be reworked on the top of `dev`.
- Key zapping is a separate problem, let's discuss it after finishing with 1 & 2.

PR #847, opened 2019-05-09
Rollback in case of view_change proof received.

vldm:
If proof was received, even without block, we should rollback

### DIFF
--- a/consensus/src/optimistic.rs
+++ b/consensus/src/optimistic.rs
@@ -29,7 +29,7 @@ use log::{debug, info};
 use std::collections::HashMap;
 use stegos_blockchain::view_changes::*;
 use stegos_blockchain::{check_supermajority, Blockchain, ChainInfo, ValidatorId};
-use stegos_crypto::hash::Hash;
+use stegos_crypto::hash::{Hash, Hashable, Hasher};
 use stegos_crypto::pbc::secure;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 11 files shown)


node/src/lib.rs
```diff
@@ -210,6 +212,7 @@ pub enum NodeMessage {
     Consensus(Vec<u8>),
     SealedBlock(Vec<u8>),
     ViewChangeMessage(Vec<u8>),
+    ViewChangeDirect(UnicastMessage),
```
rtsisyk: ViewChangeProofMessage ?

node/src/lib.rs
```diff
@@ -523,65 +503,149 @@ impl NodeService {
...
+            let local = self.chain.block_by_height(height)?;
+            let local_hash = Hash::digest(&local);
+            let local = match local {
+                Block::MicroBlock(local) => local,
+                _ => {
+                    let e = NodeBlockError::ExpectedMicroBlock(height, local_hash);
+                    panic!("{}", e);
+                }
+            };
+            ChainInfo {
```
rtsisyk: ChainInfo::from_block()

node/src/lib.rs
```diff
@@ -772,6 +826,34 @@ impl NodeService {
                     return Err(NodeBlockError::ExpectedKeyBlock(self.chain.height(), hash).into());
                 }
 
+                if self.chain.view_change() > micro_block.header.base.view_change {
```
rtsisyk: This part duplicates logic in resolve_fork(). It would be nice to have one place to handle this case.

consensus/src/optimistic.rs
```diff
@@ -39,6 +39,20 @@ pub struct ViewChangeMessage {
     pub signature: secure::Signature,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct SealedViewChangeProof {
```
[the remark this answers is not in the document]
vldm: It's used not only as message, but as internall structure for providing proofs.

node/src/lib.rs
vldm: from_block uses `view_change - 1` , but we need current view_change to compare.

node/src/lib.rs
vldm: Actualy this logic is duplicate of blockchain::validate_micro_block.

Our fork resolution makes additional checks (and somehow duplicate blockchain validation) because we need to check if block was valid before rollback.

PR #848, opened 2019-05-09
Don't send network messages until network is ready

### LINKED ISSUE #759: If node is not synchronized, it should neither send consensus messages nor send monetary blocks.
If node have outdated blockchain history, it should not flood network with consensus, or block messages.

rtsisyk:
Fix a race on start - don't send the key && micro blocks and
the view changes until network is ready.

Closes #759

### DIFF
--- a/network/src/gatekeeper/behavior.rs
+++ b/network/src/gatekeeper/behavior.rs
@@ -408,7 +408,9 @@ where
                     self.solvers.insert(peer_id.clone());
                     thread::spawn(move || {
                         let start = SystemTime::now();
+                        info!("Solving a hashcash puzzle: peer_id={:?}", peer_id);
                         let proof = hashcash::delay(p.nbits, &p.seed);
+                        info!("Solved a hashcash puzzle: peer_id={:?}", peer_id);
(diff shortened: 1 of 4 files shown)


network/src/gatekeeper/behavior.rs
```diff
@@ -408,7 +408,9 @@ where
                     self.solvers.insert(peer_id.clone());
                     thread::spawn(move || {
                         let start = SystemTime::now();
+                        info!("Solving a hashcash puzzle: peer_id={:?}", peer_id);
                         let proof = hashcash::delay(p.nbits, &p.seed);
+                        info!("Solved a hashcash puzzle: peer_id={:?}", peer_id);
```
vldm: may be add start.elapsed to output?

echupriyanov: Can you create a separate issue for this?

vldm: Why we need separete issue for "one-line" fixes, that is also related to pull request changes?

echupriyanov: Network layer fixes doesn't belong here.

PR #862, opened 2019-05-14
RestakeTransaction added as enum variant of Transaction

dbmcclain:
rebase on dev,
adopt pbc::SecretKey instead of pbc::secure::SecretKey, etc

fix comments

### DIFF
--- a/blockchain/src/validation.rs
+++ b/blockchain/src/validation.rs
@@ -28,7 +28,7 @@ use crate::error::TransactionError;
 use crate::error::{BlockError, BlockchainError};
 use crate::multisignature::check_multi_signature;
 use crate::output::Output;
-use crate::transaction::{PaymentTransaction, Transaction};
+use crate::transaction::{PaymentTransaction, RestakeTransaction, Transaction};
 use failure::Error;
 use log::*;
 use std::cmp::Ordering;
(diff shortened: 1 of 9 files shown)


blockchain/src/validation.rs
```diff
@@ -160,6 +160,148 @@ impl PaymentTransaction {
...
 }
 
+impl RestakeTransaction {
+    /// Validate the monetary balance and signature of transaction.
+    ///
+    /// # Arguments
+    ///
+    /// * - `inputs` - UTXOs referred by self.body.txins, in the same order as in self.body.txins.
+    ///
+    pub fn validate(&self, inputs: &[Output]) -> Result<StakingBalance, Error> {
```
rtsisyk: I don't like that this part partially repeats PaymentTransaction::validate(), but I don't kn

dbmcclain: Yes it causes heartburn in the Error messages too. 

When you are first constructing a RestakeTransaction, there is no transaction from which to compute the hash. But Error reporting likes to show tx and utxo hashes. So I dummied a hash of an empty string for initial creation, then use the real tx hash when later validating.

dbmcclain:
fixed create_restake_transaction, but wallet doesn't appear to know about restaking

PR #870, opened 2019-05-16
Use enum for node validation state

rtsisyk:


node/src/lib.rs
```diff
@@ -1473,39 +1527,52 @@ impl Future for NodeService {
...
-                }
-                TimerEvents::KeyBlockViewChangeTimer(_now) => {
-                    self.handle_macro_block_viewchange_timer()
-                }
-            };
-            if let Err(e) = result {
-                error!("Error: {}", e);
-            }
+        // Poll timers first.
+        let result = match &mut self.validation {
```
vldm: Do you really prefer this boilerplate code instead of macros?

You have only 4 handlers, and they are spread all over the match.

I vote for splitting this.
And minifyze count of matching arms, you can concatenate arms with same action with `|` operator
```
match &mut self.validation {
MicroBlockAuditor | MacroBlockAuditor  |
MacroBlockValidator { block_timer: BlockTimer::None,..} |
MicroBlockValidator { block_timer: BlockTimer::None,..}  => Ok(()),
// rest
}

```

this is common pattern in our program:
```
match timer.poll().unwrap() {
                Async::Ready(()) => self.handle_macro_block_viewchange_timer(),
                Async::NotReady => Ok(()),
            }
```
we probably can move this code into macros.

Additionally i didn't like the idiea of using BlockTimer for micro block, and macro block consensus, especially case when you use Dealy with deadline to current clock time.

node/src/lib.rs
```diff
@@ -1330,77 +1370,102 @@ impl NodeService {
...
     ///
-    fn create_micro_block(&mut self, proof: Option<ViewChangeProof>) -> Result<(), Error> {
-        assert!(self.consensus.is_none());
-        assert!(self.is_leader());
+    fn create_micro_block(&mut self) -> Result<(), Error> {
+        match &self.validation {
+            MicroBlockValidator { .. } => {}
+            _ => panic!("Invalid state"),
+        }
+        // assert!(clock::now().duration_since(self.last_block_clock) >= self.cfg.tx_wait_timeout);
```
vldm: could you remove this comment?

node/src/lib.rs
```diff
@@ -1330,77 +1370,102 @@ impl NodeService {
...
     }
 
     ///
     /// Create a new micro block.
     ///
-    fn create_micro_block(&mut self, proof: Option<ViewChangeProof>) -> Result<(), Error> {
-        assert!(self.consensus.is_none());
-        assert!(self.is_leader());
+    fn create_micro_block(&mut self) -> Result<(), Error> {
+        match &self.validation {
```
vldm: use assert_matches to make this contract one-line.

node/src/lib.rs
```diff
@@ -1270,50 +1270,90 @@ impl NodeService {
...
+            self.network.publish(&CONSENSUS_TOPIC, data)?;
+        }
+
+        Ok(())
+    }
+
     /// Checks if it's time to perform a view change on a micro block.
     fn handle_macro_block_viewchange_timer(&mut self) -> Result<(), Error> {
-        if self.consensus.is_none() {
+        warn!("handle_macro_block_viewchange_timer");
```
vldm: can you remove this log watning?

node/src/lib.rs
```diff
@@ -1270,50 +1270,90 @@ impl NodeService {
...
+            );
+            metrics::AUTOCOMMIT.inc();
+            // Auto-commit proposed block and send it to the network.
+            self.commit_proposed_block(block, multisig, multisigmap);
             return Ok(());
         }
 
         warn!(
-            "Timed out while waiting for a macro block: height={}",
+            "Timed out while waiting for a macro block, going to the next round: height={}",
```
vldm: should we add round to log output? The message looks like round number is awaited.

node/src/lib.rs
```diff
@@ -1270,50 +1270,90 @@ impl NodeService {
...
+    }
+
     /// Checks if it's time to perform a view change on a micro block.
     fn handle_macro_block_viewchange_timer(&mut self) -> Result<(), Error> {
-        if self.consensus.is_none() {
+        warn!("handle_macro_block_viewchange_timer");
+        assert!(clock::now().duration_since(self.last_block_clock) >= self.cfg.macro_block_timeout);
+
+        // Check that a block has been committed but haven't send by the leader.
+        let consensus = match &mut self.validation {
```
vldm: assert_matches

node/src/lib.rs
```diff
@@ -1094,167 +1045,216 @@ impl NodeService {
...
 
     ///
-    /// Returns true if current node is a leader.
+    /// Handles incoming consensus requests received from network.
     ///
-    fn is_leader(&self) -> bool {
-        self.chain.leader() == self.keys.network_pkey
-    }
+    fn handle_consensus_message(&mut self, msg: BlockConsensusMessage) -> Result<(), Error> {
+        match &mut self.validation {
```
vldm: can you simplify this match?

node/src/lib.rs
```diff
@@ -193,6 +190,10 @@ pub const VIEW_CHANGE_DIRECT: &'static str = "view_changes_direct";
 /// Topic used for sending sealed blocks.
 const SEALED_BLOCK_TOPIC: &'static str = "block";
 
+lazy_static! {
+    pub static ref LONG_TIME_AGO_CLOCK: Instant = clock::now() - Duration::from_secs(60 * 60 * 24);
```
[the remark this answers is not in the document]
rtsisyk: I removed this part.

node/src/lib.rs
rtsisyk: > I vote for splitting this.

Splitting what?

> And minifyze count of matching arms, you can concatenate arms with same action with | operator

Done:

```
        let result = match &mut self.validation {
            MicroBlockAuditor
            | MicroBlockValidator {
                block_timer: BlockTimer::None,
                ..
            }
            | MacroBlockAuditor
            | MacroBlockValidator {
                block_timer: BlockTimer::None,
                ..
            } => Ok(()),
            MicroBlockValidator {
                block_timer: BlockTimer::Propose(timer),
                ..
            } => match timer.poll().unwrap() {
                Async::Ready(()) => self.create_micro_block(),
                Async::NotReady => Ok(()),
            },
```

> we probably can move this code into macros.

I need to learn how to program using this bird language...

> Additionally i didn't like the idiea of using BlockTimer for micro block, and macro block consensus, especially case when you use Dealy with deadline to current clock time.

I don't like too. Any ideas? I tried oneshot:: and it doesn't look better (an increases match {} even more).

node/src/lib.rs
rtsisyk: Done

node/src/lib.rs

node/src/lib.rs
rtsisyk: DONE

node/src/lib.rs
rtsisyk: I added it and called it view_change. I think we should rename rounds to view_changes() to avoid confusion.

node/src/lib.rs
rtsisyk: Is not applicable here.

node/src/lib.rs

node/src/lib.rs
rtsisyk: timer.poll() is just 4 lines of code, 16 total. To call macro, you need at least 1 line, that leaves 12 lines for macro. `poll_timer!()` macro was 11 lines long. That means, that macro saves just one line, but degrades readability significantly.

```
match timer.poll().unwrap() {
                Async::Ready(()) => self.create_micro_block(),
                Async::NotReady => Ok(()),
            },
```

PR #871, opened 2019-05-16
AONT Encrypt / Decrypt provide in stegos_crypto

dbmcclain:
Provide AONT Encrypt / Decrypt in stegos_crypto
(will be used for serialization of SecretKeys)


dbmcclain:
Added encrypted serialization for SecretKey

PR #885, opened 2019-05-18
Dm wipe after use

dbmcclain:
initial commit - minimal changes with by-val semantics


crypto/src/curve1174/lev32.rs
```diff
@@ -27,11 +27,29 @@ use std::cmp::Ordering;
 // -----------------------------------------------------------------
 // type Lev32 represents a 256-bit bignum as a little-endian 32-byte vector
 
-#[derive(Copy, Clone, Eq, PartialEq)]
-pub struct Lev32(pub [u8; 32]);
+#[derive(Clone)]
+pub struct Lev32(pub [u8; 32], pub bool);
```
vldm: why you need this secure clearing, to be conditional?

vldm: Additionally it's still up to LLVM to copy data, that is saved on stack, are you sure that this patch will add some security, to our system?

dbmcclain: We need conditional because there is one body of code, via Fields macro, to generate Fq and Fr items. Only Fr will be at risk and needs clearing, and only then, if the value has been derived from secret information. Small constants like 2, don't need to be cleared after use.

As for LLVM, we are ensuring that data gets zeroed by calling out to external C stub after zeroing. Ostensibly that C routine could do anything, but it only does a return - just enough to force the Rust compiler to follow through and make that call with "valid" data, which is zero.

dbmcclain:
Speed up BulletProof validation: from 81ms down to 23ms

crypto/protos/crypto.proto
```diff
@@ -6,7 +6,8 @@ message Pt {
 }
 
 message Fr {
-    bytes data = 1;
+    bool wau = 1;
```
rtsisyk: Why this flag is needed? It breaks genesis.

dbmcclain: It is needed for restore from deserialization. I have to know whether the value (just data bytes) is an Fr value directly? or AONT cloaked. Just regenerate the genesis package.

PR #900, opened 2019-05-22
Remove KeyChain from the most places

rtsisyk:


blockchain/src/test.rs
```diff
@@ -0,0 +1,287 @@
...
+    timestamp: SystemTime,
+) -> (Vec<KeyChain>, MacroBlock) {
+    let mut keychains = Vec::with_capacity(num_nodes);
+    for _i in 0..num_nodes {
+        let keychain = KeyChain::new();
+        keychains.push(keychain);
+    }
+    let mut stakes = Vec::with_capacity(num_nodes);
+    for i in 0..num_nodes {
+        let stake_def = StakeDef {
```
vldm: I didn't like the idea of creating struct just for one case.
Could we create stakeoutput there?

keychain/src/input.rs
```diff
@@ -64,15 +68,15 @@ fn read_recovery_from_file(recovery_file: &str) -> Result<curve1174::SecretKey,
     Ok(recovery_to_wallet_skey(&recovery)?)
 }
 
-pub(crate) fn read_recovery(recovery_file: &str) -> Result<curve1174::SecretKey, KeyError> {
-    if recovery_file == "-" && atty::is(atty::Stream::Stdin) {
+pub fn read_recovery(recovery_file: &str) -> Result<curve1174::SecretKey, KeyError> {
+    if is_input_interactive(recovery_file) {
```
vldm: i dont get it. Why you check if input interactive based on file name?

src/console.rs
```diff
@@ -265,42 +244,28 @@ impl ConsoleService {
...
-            " - generator start LIST_OF_WALLETS_ADDRESSES - will start generator, \
-             or add new wallet keys to generator list."
-        );
-        println!(" - generator stop will stop generator");
-        println!();
-    }
-
     /// Called when line is typed on standard input.
-    fn on_input(&mut self, msg: &str) -> bool {
+    fn on_input(&mut self, msg: &str) -> Result<bool, Error> {
```
vldm: Why this is changed?

blockchain/src/test.rs
rtsisyk: I will try to replace StakeDef with created 'StakeOutput'.

keychain/src/input.rs
rtsisyk: There are two configuration options in our config file:

 - `password_file` - specifies a path to file with saved password for keys.
 - `recovery_file` - specifies a path to file with 24-word recovery.

When value is empty or `-` the password or recovery is read from stdin. For example,

`stegos --recovery-file -` will start interactive recovery.

src/console.rs
rtsisyk: After adding password to API the most functions can return an error on attempt to read the password from the disk. I replaces all unwraps() with `?` and added handling of errors on the upper level.

keychain/src/input.rs
vldm: Could we not introduce specific value for filename, and use `Option` in rust to check if file was specified?

also could we check if tty is interactive? And write error if file was not specified and tty is not interactive?

rtsisyk: > 

Could we not introduce specific value for filename, and use Option in rust to check if file was specified?

> also could we check if tty is interactive? And write error if file was not specified and tty is not interactive?

Ok, I will change it during the further refactoring of wallet.

PR #915, opened 2019-05-23
Add r-value return from AES encrypt, insert H(P_spender) into encrypted payload

dbmcclain:
initial commit

### DIFF
--- a/blockchain/src/genesis.rs
+++ b/blockchain/src/genesis.rs
@@ -59,6 +59,9 @@ pub fn genesis(
         // Genesis has one PaymentOutput + N * StakeOutput, where N is the number of validators.
         //
 
+        // TODO: use correct spender pubkey
+        let spender_pkey = keychains[0].wallet_pkey;
+
         // Node #1 receives all moneys except stakes.
         // All nodes gets `stake` money staked.
         //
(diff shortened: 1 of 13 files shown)


dbmcclain:
AES Encrypt now returns encrypted payload struct, plus the secret r-value used for keying
Hash of spender pubkey now stored in encrypted payload
Generating PaymentOutput now requires pubkey of spender, will be hashed and stored in encrypted payload.
New PaymentOutput functions return the UTXO, the gamma, and the r-value

dbmcclain:
Added function to PaymentOutput:
verify_proof_of_payment(&UTXO, &spender_pkey, &r-value) -> Result<amount: i64, Error>

dbmcclain:
fix mistaken keying on encrypted payload:
key = H(r * s *G) = H(r * P)
for P = uncloaked recipient key
so verify_proof_of_payment() needs UTXO, spender_pkey, r-value, uncloaked_recipient_pkey.

blockchain/src/genesis.rs
```diff
@@ -59,6 +59,9 @@ pub fn genesis(
         // Genesis has one PaymentOutput + N * StakeOutput, where N is the number of validators.
         //
 
+        // TODO: use correct spender pubkey
```
rtsisyk: In genesis node[0] sends money to itself. There is no "correct" spender pubkey.

dbmcclain:
> I have a conceptual question here - how storing of sender's public key in EncryptedPayload can prove that this UTXO was actually created by that wallet?

yes you are correct - we need a signature instead of key-hash

dbmcclain:
just sync'd with dev branch

rtsisyk:
I will review it after v0.8 release.

rtsisyk:
Superseded by https://github.com/stegos/stegos/pull/970

PR #919, opened 2019-05-24
Fill up macro blocks with real content

rtsisyk:
Add inputs && outputs from entire epoch.

Needed for new MacroBlocks


rtsisyk:
Replaced by #932.

PR #924, opened 2019-05-27
Reconnect failed mutual connection attempts

echupriyanov:
... also simplified Pubsub/Gatekeeper negotiations


network/src/gatekeeper/behavior.rs
```diff
@@ -609,3 +653,21 @@ pub enum ListenerPeerState {
     WaitingProof,
     Unlocked,
 }
+
+// Trait for debugging external types
+
+trait StegosDisplay {
```
rtsisyk: OMG :)

PR #931, opened 2019-05-30
Prepare for new Macro Blocks

rtsisyk:
- Rename MVCC `version` to `lsn` to avoid confusion with Block `version`.
- Move handling of InvalidViewChange from apply_micro_block() to handle_block()
- Refactor validate_micro_block()

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -122,11 +122,21 @@ pub(crate) struct Balance {
     pub block_reward: i64,
 }
 
-type BlockByHashMap = MultiVersionedMap<Hash, u64, u64>;
-type OutputByHashMap = MultiVersionedMap<Hash, OutputKey, u64>;
-type BalanceMap = MultiVersionedMap<(), Balance, u64>;
+#[derive(Debug, Default, Clone, Copy, PartialOrd, Ord, PartialEq, Eq)]
+pub(crate) struct LSN(u64); // use `struct` to disable explicit casts.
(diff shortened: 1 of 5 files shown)


blockchain/src/blockchain.rs
```diff
@@ -122,11 +122,21 @@ pub(crate) struct Balance {
...
 
-type BlockByHashMap = MultiVersionedMap<Hash, u64, u64>;
-type OutputByHashMap = MultiVersionedMap<Hash, OutputKey, u64>;
-type BalanceMap = MultiVersionedMap<(), Balance, u64>;
+#[derive(Debug, Default, Clone, Copy, PartialOrd, Ord, PartialEq, Eq)]
+pub(crate) struct LSN(u64); // use `struct` to disable explicit casts.
+const INITIAL_LSN: LSN = LSN(0);
+
+/// Create LSN for MultiVersionedMap from height.
+fn lsn_for_height(height: u64) -> LSN {
```
vldm: 1) In rust naming convention: https://doc.rust-lang.org/1.0.0/style/style/naming/README.html

`acronyms count as one word` so we should name it `Lsn` rather than `LSN`

2) Also could you also make this method part of structure implementation?
3) And may be move it into mvcc.rs ? And make it as default type of version.

rtsisyk: 1. Ok, let's rename it later.
2. Ok
3. mvcc.rs know nothing about blockchain's height.

PR #954, opened 2019-06-04
Make utxos lockable.

vldm:
Add possibility to lock utxos from spending.


blockchain/src/output.rs
```diff
@@ -383,7 +397,17 @@ impl PaymentOutput {
     /// Create a new PaymentOutput.
     pub fn new(recipient_pkey: &PublicKey, amount: i64) -> Result<(Self, Fr), BlockchainError> {
         let data = PaymentPayloadData::Comment(String::new());
-        Self::with_payload(recipient_pkey, amount, data)
+        Self::with_payload(recipient_pkey, amount, data, None)
+    }
+
+    /// Create a new PaymentOutput with lock.
+    pub fn new_locked(
```
rtsisyk: What is about locked + comment case?

blockchain/src/protos.rs
```diff
@@ -53,17 +53,33 @@ impl ProtoConvert for PaymentOutput {
         proto.set_cloaking_hint(self.cloaking_hint.into_proto());
         proto.set_proof(self.proof.into_proto());
         proto.set_payload(self.payload.into_proto());
+        let timestamp = if let Some(locked_timestamp) = self.locked_timestamp {
```
rtsisyk: Please extract serialization of SystemTime into stegos_serialization. I understand that it was added before, but now we have the same code in multiple places. It would be better to extract it into a function.

node/src/lib.rs
```diff
@@ -161,6 +161,7 @@ pub struct BlockAdded {
 #[derive(Clone, Debug, Serialize)]
 pub struct EpochChanged {
     pub epoch: u64,
+    pub last_block_time: SystemTime,
```
rtsisyk: last_macro_block_timestamp ?

src/console.rs
```diff
@@ -182,8 +187,10 @@ impl ConsoleService {
 
     fn help() {
         println!("Usage:");
-        println!("pay WALLET_PUBKEY AMOUNT [/public] [COMMENT] - send money");
-        println!("spay WALLET_PUBKEY AMOUNT [COMMENT] - send money using ValueShuffle");
+        println!("pay WALLET_PUBKEY AMOUNT [/public] [COMMENT] [/lock duration] - send money");
```
rtsisyk: It would be better to group all modifers, i.e. /public and /lock should be together.

wallet/src/lib.rs
```diff
@@ -164,6 +191,9 @@ pub struct WalletService {
     /// Lifetime of stake.
     stake_epochs: u64,
 
+    /// Time of last macro block.
+    last_block_time: SystemTime,
```
rtsisyk: last_macro_block_timestamp

blockchain/src/output.rs
vldm: With payload is exist for this.

PR #965, opened 2019-06-06
Payment certificate api

vldm:
Feature that i decide to move to separate task: 

1 Rewrite transaction interests, to use link to log, and make it recoverable after restart
1.2 Also update status of transaction on tx commit or cancel.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -33,12 +33,14 @@ use std::fmt;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::thread;
-use std::time::{Duration, SystemTime};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 pub use stegos_api::url;
 use stegos_api::*;
 use stegos_crypto::curve1174::PublicKey;
 use stegos_crypto::pbc;
 use stegos_keychain::input;
+
+const CONSOLE_HISTORY_LIMIT: u64 = 50;
 
 // ----------------------------------------------------------------
 // Public API.
@@ -154,6 +156,7 @@ impl ConsoleService {
         println!("show keys - print keys");
         println!("show balance - print balance");
         println!("show utxo - print unspent outputs");
+        println!("show history [STARTING DATE] - print history since date");
         println!("show election - print leader election state");
         println!("show escrow - print escrow");
         println!("show recovery - print recovery information");
@@ -575,6 +578,15 @@ impl ConsoleService {
             self.send_node_request(request)?
         } else if msg == "show utxo" {
             let request = WalletRequest::UnspentInfo {};
+            self.send_wallet_request(request)?
+        } else if msg.starts_with("show history") {
+            let arg = &msg[12..];
+
+            let starting_from = humantime::parse_rfc3339(arg).unwrap_or(UNIX_EPOCH);
+            let request = WalletRequest::HistoryInfo {
+                starting_from,
+                limit: CONSOLE_HISTORY_LIMIT,
+            };
             self.send_wallet_request(request)?
         } else if msg == "show recovery" {
             let password = input::read_password_from_stdin(false)?;
--- a/wallet/protos/wallet_log.proto
+++ b/wallet/protos/wallet_log.proto
@@ -1,0 +1,62 @@
+syntax = "proto3";
+package stegos.wallet_log;
+
+import "crypto.proto";
+import "blockchain.proto";
+
+message PaymentPayload {
+    oneof data {
+        string comment = 1;
+        bytes hash = 2;
+    }
+}
+
+message PaymentValue {
+    stegos.blockchain.PaymentOutput output = 1;
+    int64 amount = 2;
+    string comment = 3;
+}
+
+// Possible outputs
+// (Copy of stegos.blockchain.Output, but without stake,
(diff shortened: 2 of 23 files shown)


wallet/src/api.rs
```diff
@@ -110,6 +144,10 @@ pub enum WalletRequest {
 #[serde(tag = "response")]
 #[serde(rename_all = "snake_case")]
 pub enum WalletResponse {
+    TransactionCreatedWithCertificate {
+        tx_hash: Hash,
+        info: PaymentTransactionInfo,
```
rtsisyk: Why just don't add rvalue as Option<> to existing TransactionCreated?

src/console.rs
```diff
@@ -564,6 +564,12 @@ impl ConsoleService {
         } else if msg == "show utxo" {
             let request = WalletRequest::UnspentInfo {};
             self.wallet_response = Some(self.wallet.request(request));
+        } else if msg == "show history" {
+            let request = WalletRequest::HistoryInfo {
+                offset: 0,
```
rtsisyk: Offset/Limit approach is slow, because you need to scroll entire table to skip `offset` records. It would be better to replace use optional last know key instead `offset`.  Even better for this particular table is to ask for [begin, end) timestamps.

We also need to understand how to support for this feature in console...

wallet/protos/wallet_log.proto
```diff
@@ -0,0 +1,65 @@
...
+    stegos.crypto.Fr rvalue = 3;
+    int64 amount = 4;
+}
+
+// Information about created transactions
+message PaymentTransactionValue {
+    stegos.blockchain.PaymentTransaction tx = 1;
+    repeated PaymentCertificate certificates = 2;
+    /// source PublicKey, save in case of changing.
+    stegos.crypto.PublicKey source = 3;
```
rtsisyk: I don't think that it is possible to change public key in this system. Let's ask David.

wallet/protos/wallet_log.proto
```diff
@@ -0,0 +1,65 @@
...
+import "blockchain.proto";
+
+message PaymentPayload {
+    oneof data {
+        string comment = 1;
+        bytes hash = 2;
+    }
+}
+
+message PaymentValue {
```
rtsisyk: Why don't store original UTXO here?  We lose some information by replacing PaymentOutput with PaymentValue. PaymentValue initially was added to store decrypted balance to speed up find_utxo(). Now I think that this optimization doesn't make sense, because AES is fast enough and you can decode UTXO on the fly.

wallet/src/api.rs
```diff
@@ -35,6 +35,27 @@ use stegos_crypto::hash::Hash;
...
+    },
+    Outgoing {
+        timestamp: SystemTime,
+        tx: PaymentTransactionValue,
+    }
+}
+
+#[derive(Serialize, Clone, Debug, PartialEq, Eq)]
+pub enum OutputInfo {
+    Cloaked(PaymentInfo),
```
rtsisyk: I don't like this idea of different names for the same thing. We should either name this variant `Payment` or rename `PaymentOutput` to `CloakedPayment`.

wallet/protos/wallet_log.proto
rtsisyk: Hmm.. I see the problem. The full UTXO should be stored on the disk, but for API a decoded version is needed....

rtsisyk: What do you think about storing original UTXO/Transactions on the disk, but adding PaymentInfo + TransactionInfo for API. I see some benefits here:

- UTXO will be stored in encrypted form on the disk.
- All information is preserved in original form.
- API will receive only required information.
- The output of API can be changed without changing the disk structure.

wallet/src/api.rs
```diff
@@ -129,6 +154,10 @@ pub enum WalletRequest {
 #[serde(tag = "response")]
 #[serde(rename_all = "snake_case")]
 pub enum WalletResponse {
+    TransactionCreatedWithCertificate {
+        tx_hash: Hash,
+        info: PaymentTransactionValue,
```
rtsisyk: Above I see that tx can generate multiple certificates, but here is just one certificate is available. Probably it should be Vec<PaymentCertificate> instead. I also suggest to merge `TransactionCreatedWithCertificate` into `TransactionCreated` to simplify the things.

wallet/src/protos.rs
```diff
@@ -0,0 +1,219 @@
+//! message.rs - ValueShuffle Protobuf Encoding.
```
rtsisyk: Doesn't look like ValueShuffle.

wallet/src/storage.rs
```diff
@@ -0,0 +1,381 @@
...
+            .take(limit as usize)
+    }
+
+    //
+    // Internal api.
+    //
+
+
+    /// Convert idx to bytearray.
+    fn key(idx: u64) -> [u8; 8] {
```
rtsisyk: Probably timestamp should be better than a surrogate primary key, because it will allow fast [begin_timestamp, end_timestamp) scan for the history. But please ensure that timestamp is always increasing (for example, by using gettimeofday()-based function, which always produces increasing timestamp or by adding some extra milliseconds manually on conflict).

PR #970, opened 2019-06-07
Add r-value return from AES encription.

vldm:
This is recover of Davids patch   #915.

I just changed a bit format of encrypted payload proposed by David, to store signature rather then public key of sender.

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -282,7 +282,11 @@ impl Blockchain {
     // Recovery.
     //----------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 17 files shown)


blockchain/src/blockchain.rs
```diff
@@ -764,6 +768,7 @@ impl Blockchain {
     pub fn create_macro_block(
         &self,
         view_change: u32,
+        spender_key: &curve1174::SecretKey,
```
rtsisyk: Node should work without wallet' secret key. Payment Certificates are optional and should not be used for reward/fee UTXO.

PR #974, opened 2019-06-11
Websocket API message encryption implemented

### LINKED ISSUE #946: WebSocket API Authorization
Add authorization to WebSocket API

echupriyanov:
* All messages are encrypted with AES256 encryption and encoded in base64 to transmit as WebSocket text message for easier debugging
* API key is read from file, which can be configured in config file (default: api_token.txt)

Closes #946

### DIFF
--- a/api/src/crypto.rs
+++ b/api/src/crypto.rs
@@ -1,0 +1,74 @@
+//! WebSocket API - Messages encryption.
+
+//
+// MIT License
+//
+// Copyright (c) 2019 Stegos AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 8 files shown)


api/src/error.rs
```diff
@@ -0,0 +1,38 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+use failure::Fail;
+
+#[derive(Debug, Fail)]
+pub enum KeyError {
```
rtsisyk: Looks liks a copy-paste from stegos_keychain.

api/src/crypto.rs
```diff
@@ -0,0 +1,73 @@
...
+// SOFTWARE.
+
+use crypto::aes::{self, KeySize};
+use crypto::symmetriccipher::SynchronousStreamCipher;
+use rand::{thread_rng, RngCore};
+use std::iter::repeat;
+
+// Encrypts the plaintext with given 32-byte key
+// returns encrypted payload with 16-byte IV prepended
+pub fn encrypt(key: &[u8], plaintext: &[u8]) -> Vec<u8> {
```
rtsisyk: @dbmcclain could you please review this part?

dbmcclain: Looks okay to me. I see you using a default nonce of all zeros, and you write that nonce out as the first 16 bytes of the encrypted payload. 

Potentially unnecessary to write the nonce, but maybe you have plans to use arbitrary nonces in the future?

echupriyanov: No, no, no...
```
let mut nonce: Vec<u8> = repeat(0u8).take(16).collect();
gen.fill_bytes(&mut nonce[..]);
```
fills `nonce` with random stuff

api/src/error.rs
echupriyanov: Mostly it, with minor difference with decoding errors

api/src/crypto.rs
dbmcclain: yes - good

api/src/error.rs
dbmcclain: LGTM

PR #976, opened 2019-06-11
Refactor stake validation

rtsisyk:
Move stake validation into one place.

This patch is pure refactoring.

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -574,14 +574,17 @@ impl Blockchain {
         self.election_result.random.rand
     }
 
-    ///
-    /// Get staked value for validator.
-    ///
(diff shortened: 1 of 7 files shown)


blockchain/src/blockchain.rs
```diff
@@ -593,6 +596,18 @@ impl Blockchain {
     pub(crate) fn staker_outputs(&self, validator_pkey: &pbc::PublicKey) -> (Vec<Hash>, i64) {
         self.escrow.staker_outputs(validator_pkey, self.epoch)
     }
+
+    ///
+    /// Return a wallet key by network key.
+    ///
+    #[inline]
+    pub(crate) fn wallet_by_network_key(
```
vldm: I am not sure that this name is better, because currently escrow only handle stakers keys. So we probably should add some docs info about this?

PR #984, opened 2019-06-14
Prepare for remote console

rtsisyk:
Please review individual commits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,6 +2630,7 @@ dependencies = [
  "serde_derive 1.0.92 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.39 (registry+https://github.com/rust-lang/crates.io-index)",
  "stegos_crypto 0.2.0",
+ "stegos_network 0.4.0",
  "stegos_node 0.2.0",
  "stegos_wallet 0.2.0",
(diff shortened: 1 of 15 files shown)


rtsisyk:
I haven't touched `net send` and `net publish` in this patch. The new API allow to subscribe to any particular unicast/broadcast topic. It can be used for debugging purposes. For example, someone can subscribe to `consensus` topic and start auditing messages.  I plan add `net subscribe broadcast TOPIC` and `net subscribe unicast TOPIC` to Console after switching to Public API. I don't see any reason to don't allow arbitrary topics in Console. Of course, I can subscribe Console to "console" topic by default.

echupriyanov:
You can't subscribe to arbitrary topics in broadcast communications during runtime. To propagate message through broadcast, all involved nodes should be subscribed to the same topic. It just makes no sense to subscribe to topic on only one or couple of nodes. Intermediate nodes just wouldn't pass such messages.

rtsisyk:
Yes, but you can subscribe to existing topics, e.g. "block" or "consensus" via API.

echupriyanov:
Yes, exactly. That's why I suggest to subscribe WebSocket/Node service to common topic exclusively to be used by console.

rtsisyk:
Agreed to enable `console` topic by default.

rtsisyk:
Replaced by #981

PR #990, opened 2019-06-17
Add support for variable tx fees

rtsisyk:
- Add support for `payment_fee` in API.
- Allow user to specify fee in Console.
- Sort mempool transactions by fee.


node/src/mempool.rs
```diff
@@ -171,18 +170,32 @@ impl Mempool {
         let seed = mix(last_random, view_change);
         let random = pbc::make_VRF(network_skey, &seed);
 
+        // Sort transactions by fee.
+        // RestakeTransactions have high priority.
+        let mut pool: Vec<(Hash, i64)> = self
```
vldm: May be it would good to just implement ord manually?

PR #991, opened 2019-06-17
Implement changing of wallet password

rtsisyk:


rtsisyk:
> We should remove password protection from the network keys on disk, since these keys are ephemeral and doesn't require hard protection

OK

echupriyanov:
Looks good now

PR #1019, opened 2019-06-20
Refactor dependencies

### LINKED ISSUE #1018: Use only necessary parts of `rust-libp2p`
Instead of depending on whole `libp2p` crate, only import parts we are actually depend on

echupriyanov:
* libp2p switched to use separate sub-crates
* same for tokio
* libp2p bumped to v0.9.1
* replaced secp256k1 for secio/peer_id with ed25519

Closes #1018


Cargo.lock.orig
```diff
@@ -0,0 +1,3857 @@
+# This file is automatically @generated by Cargo.
```
rtsisyk: Please remove

PR #1027, opened 2019-06-24
Update benchmarks

rtsisyk:


rtsisyk:
Fixed negative bulletproofs.

PR #1045, opened 2019-06-26
Remove low-level Hash functions, remove extraneous PBC compress() fun…

dbmcclain:
Removed extraneous PBC compress() function for both G1 and G2
Improve pbc::PublicKey decompress() to avoid spurious reconstruction of PublicKey if already decompressed

In SCC:
Remove low-level Ristretto component Hash. This changes hashing of Fr, SecretKey, SchnorrSignature, and encrypted payload hashes. But no change to hash of PublicKeys.
Improve scc::PublicKey decompress() to avoid spurious reconstruction of PublicKey if already decompressed

No key regen needed for unit tests

### DIFF
--- a/crypto/src/pbc/mod.rs
+++ b/crypto/src/pbc/mod.rs
@@ -111,14 +111,6 @@ impl G1 {
         }
     }
 
-    pub fn compress(&self) -> Self {
-        match self {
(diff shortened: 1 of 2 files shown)


rtsisyk:
David: this patch knowingly changes hashing, but not on PublicKeys

PR #1050, opened 2019-06-27
add test for constant time SCC verification

dbmcclain:
Adds a unit test to check for constant time SCC performance

### DIFF
--- a/crypto/src/scc/mod.rs
+++ b/crypto/src/scc/mod.rs
@@ -1201,4 +1201,32 @@ pub mod tests {
         let deserialized: PublicKey = serde_json::from_str(&serialized).unwrap();
         assert_eq!(pkey, deserialized);
     }
+
+    #[test]
+    fn check_constant_time() {
+        use std::time::SystemTime;
+        let skey = SecretKey::from(Fr::one());
+        let kval = Fr::one();
+        let sumk = Pt::inf();
+        let sumpkey = Pt::inf();
+        let h = Hash::from_str("Testing");
+        let niter = 100_000;
+        let start = SystemTime::now();
+        for _ in 0..niter {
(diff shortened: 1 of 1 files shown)


crypto/src/scc/mod.rs
```diff
@@ -1201,4 +1201,32 @@ pub mod tests {
...
+        let skey = SecretKey::from(
+            Fr::try_from_hex("0FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF")
+                .expect("ok)"),
+        );
+        let start = SystemTime::now();
+        for _ in 0..niter {
+            sign_hash_with_kval(&h, &skey, &kval, &sumk, &sumpkey);
+        }
+        let timing = start.elapsed().unwrap();
+        println!("Const Time (1111...) = {:?}", timing / niter);
```
rtsisyk: What is checked by this test?

dbmcclain: This tests the timing constancy of forming a Schnorr Signature by using a secret key with only  bit set, versus another secret key with a large number of bits set. The printout shows that, to within typical measurement variances, the timing is the same.

Specifically I test the formation of `u = k + H(k*G, s*G, h_msg) * s`. I'm looking at that last multiply by secret key s, for fixed values of k, and h_msg.

dbmcclain: I guess I'm also seeing the effects of that `s*G` inside the Hash value. So I get to see two secret key multiplications, which should serve to double any timing variance with number of bits set in secret key.

Constant-time arithmetic should be insensitive to the number of set bits in the secret key. And running this test shows that is so.

PR #1068, opened 2019-07-04
Move automatic re-staking from Wallet to Node

rtsisyk:


rtsisyk:
Reworked smoke_test() to test re-staking.

PR #1073, opened 2019-07-05
VDFMon - Support VDF's for constant block production rate

dbmcclain:
add vdf folder

cargo fmt

Added code for VDFMon objects
analyze inter-block durations
compute VDF complexity levels
form a VDF proof
validate a VDF proof and update to new complexity level

fix code errors

ignore timing tests


rtsisyk:
Replaced by #1110

PR #1091, opened 2019-07-11
Make snowball return tx created response, when transaction formed.

vldm:
Also convert SnowballCreated to notification message.


wallet/src/lib.rs
```diff
@@ -963,6 +952,30 @@ impl Future for UnsealedAccountService {
                                     error: format!("{}", e),
                                 },
                             },
+                            AccountRequest::SecurePayment {
```
rtsisyk: I didn't get idea why this block has been moved.

PR #1114, opened 2019-07-15
Limit number of input UTXOs in composed transactions

### LINKED ISSUE #1113: Update transaction create with limit on input UTXOs
* Limit inputs in created UTXO by settings from Node's config
* Use largest UTXO to get necessary payment amount, than fill inputs with smallest UTXOs.

echupriyanov:
Also, tries to fill up inputs with smallest UTXOs to burn 'em.

Closes #1113

### DIFF
--- a/wallet/src/change.rs
+++ b/wallet/src/change.rs
@@ -23,6 +23,7 @@
 
 use crate::error::*;
 use log::trace;
+use std::collections::VecDeque;
 use stegos_blockchain::Timestamp;
 
 /// Find appropriate inputs.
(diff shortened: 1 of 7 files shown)


wallet/src/change.rs
```diff
@@ -66,14 +68,15 @@ where
 
     // Sort in ascending order to eliminate as much outputs as possible
     sorted.sort_by_key(|(amount, _output)| *amount);
+    let mut inputs: VecDeque<(i64, &T)> = VecDeque::from(sorted);
 
     // Try to spend without a change.
     let mut spent: Vec<&T> = Vec::new();
     let mut change: i64 = sum + fee;
-    for (amount, output) in sorted.iter() {
+    for (amount, output) in inputs.iter() {
```
vldm: Maybe remove this code, it looks redurant?

echupriyanov: The first case uses regular fee, which is less than change fee. Maybe this can be optimized. But I'm not sure it will gain us much

Issue #1124, opened 2019-07-17
Make tests reproducible with fixed seed.


vldm:
The idea is following:

1. Add a pRNG object into `Sandbox`.
This pRNG should be inited with some random seed at sandbox startup.
This initial seed should be printed into console.
2) There should be api, that will generate pseudo random `Hash` value, from pRNG.
3) All `genesis` and `keys` creation in tests, should be used with `make_deterministic` . 
4) There also should be possible to reproduce some test with specific seed.

PR #1143, opened 2019-07-19
Add seed for generated keys in tests.

vldm:
Closes: #1124

Also need to update rand library, to use latest `rand_core`.

Tested on different leaders in microblock - it works.
But we still creating Payments output with random value, and probably need to make them more deterministic.


blockchain/src/test.rs
```diff
@@ -94,6 +99,7 @@ pub fn fake_genesis(
 
     // Create an initial payment.
     assert!(payout > 0);
+    //TODO: Should we also create outputs in predictable way?
```
rtsisyk: @dbmcclain it would be nice to have make_deterministic_bulletproof or something like that in order to make unit tests predictable. Our idea is to generate some random seed value before starting the test system and print this seed to console. This seed value can be used to re-produce the test in case of failure. Currently we struggle to re-produce failures of these randomized tests.

PR #1171, opened 2019-07-26
ValueShuffle enhancement

### LINKED ISSUE #1092: Limit number of Snowball restarts
Currently, in case of failure Snowball tries to restart session. This can lead to infinite loop within Snowball.

We need to limit number of restart attempts, and allow Snowball return Error results, and not only successfully created super-transaction.

echupriyanov:
Now VS can result in 3 ways:
* SignedTransaction(tx, leader) - the same it was before
* UnsignedTransaction(tx) - transaction formed, but not all signatures were collected
* Failure(Error) - fatal failure during protocol run

Also, added `terminate` method to VS, to reset its state and stop further participation in current session (if any)

Closes #1092

### DIFF
--- a/wallet/src/lib.rs
+++ b/wallet/src/lib.rs
@@ -66,7 +66,7 @@ use tokio_timer::{clock, Interval};
 
 const STAKE_FEE: i64 = 0;
 const RESEND_TX_INTERVAL: Duration = Duration::from_secs(2 * 60);
-const PENDING_UTXO_TIME: Duration = Duration::from_secs(2 * 60);
+const PENDING_UTXO_TIME: Duration = Duration::from_secs(5 * 60);
 const CHECK_PENDING_UTXO: Duration = Duration::from_secs(10);
 
 ///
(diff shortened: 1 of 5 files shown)


wallet/src/lib.rs
```diff
@@ -856,20 +856,23 @@ impl Future for UnsealedAccountService {
...
+                                let _ =
+                                    response_sender.send(AccountResponse::TransactionCreated(tx));
+                            }
+                            Err(e) => {
+                                error!("Error during processing valueshuffle transaction = {}", e)
+                            }
                         }
                     }
-                }
+                    ValueShuffleOutput::UnsignedTransaction(_tx) => unimplemented!(),
```
vldm: probably we should replace `unimplemented!` macro with

```
error!("Unimplemented UnsignedTransaction");
```

wallet/src/valueshuffle/mod.rs
```diff
@@ -74,6 +74,7 @@
 // ========================================================================
 
 #![allow(non_snake_case)]
+#![allow(warnings)]
```
vldm: Why you allowing warnings?
Can you do it more granular.
For example, if you didn't use only `ValueShuffleOutput::SignedTransaction`, mark `#[allow(warning)]` this variant.

echupriyanov: It complies on unused `terminate` method. we can remove `allow(warnings)` as soon as this method will be used

wallet/src/lib.rs
echupriyanov: This is very temporal placeholder, to be filled, when UTXO and payment tracking will be implemented

PR #1175, opened 2019-07-26
QueryShuffle - Pared Down ValueShuffle Protocol for Anonymous Database Queries

dbmcclain:
initial commit

### DIFF
--- a/crypto/protos/crypto.proto
+++ b/crypto/protos/crypto.proto
@@ -58,11 +58,12 @@ message BulletProof {
 }
 
 message DiceMixParticipantID {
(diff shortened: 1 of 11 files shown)


dbmcclain:
not yet ready for review - just here to get things started, and so that I could rebase off of latest dev

dbmcclain:
I think this is ready for initial review. It currently exists as a protocol service module in the system, but nobody is calling it to do anything yet. 

There remains a need to provide a facilitator logic and server database lookup, and wallet responses to query replies. But this central service allows wallets to submit UTXO lookup requests from an anonymous query pool.

PR #1190, opened 2019-07-30
fix ValueShuffle 

dbmcclain:
- should not enter dc_decode when number of participants falls below expected number from earlier matrix construction

It is apparent that checking self.pending_participants is not sufficient to ensure that original number of participants remain. So we were calling DiceMix::dc_decode() with inconsistent numbers of matrices when there had been drop out participants. Not all dropouts were tallied by self.pending_participants. 

(Perhaps participants who supplied invalid messages were also removed from pending_participants?) 

This patch to ValueShuffle should prevent a large number of panics that have been experienced recently. No material change to DiceMix (just some test code at the bottom, and removed yesterday's patch to count MIA's). This time if DiceMix panics then something is seriously wrong.

### DIFF
--- a/crypto/src/dicemix/mod.rs
+++ b/crypto/src/dicemix/mod.rs
@@ -222,6 +222,7 @@ pub fn split_message(msg: &[u8], max_cols: Option<usize>) -> DcRow {
                 Some(ncmax) => assert!(ncols <= ncmax, "Message requires too many cols"),
             }
             stuff_slot(&mut seg, &msg, NPREF, offs, NCHUNK);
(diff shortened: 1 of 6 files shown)


dbmcclain:
Please don't use yet - there are still glaring problems with it...

dbmcclain:
Serious rewrite of ValueShuffle internal logic, 
Fix of DiceMix errors in shared scalar values
This can't be any worse than what we already had...

echupriyanov:
Is it ready for testing?

dbmcclain:
yes

PR #1213, opened 2019-08-05
Fix is_change in incomming history entry.

vldm:


wallet/src/storage.rs
```diff
@@ -26,7 +26,7 @@
 use crate::api::*;
 use byteorder::{BigEndian, ByteOrder};
 use failure::{bail, Error};
-use log::{debug, trace};
+use log::{debug, error, trace};
```
rtsisyk: Please `use log::*` instead.

PR #1229, opened 2019-08-08
Refactor TransactionPoolService

rtsisyk:


node/src/lib.rs
```diff
@@ -1895,10 +1909,12 @@ impl Future for NodeService {
             error!("Error: {}", e);
         }
 
-        match self.txpool_service.poll().expect("txpool working") {
-            Async::Ready(_) => unreachable!(),
-            Async::NotReady => {}
-        };
+        if let Some(ref mut txpool_service) = &mut self.txpool_service {
```
vldm: you can remove if let, and `poll` `Option<Future>` directly

PR #1247, opened 2019-08-13
Rework network configuration

rtsisyk:
- Replace `bind_ip` and `bind_port` with `endpoint`.
- Add `STEGOS_NETWORK_ENDPOINT` environment variable.
- Replace `advertised_addresses` with `public_endpoint`.
- Add `STEGOS_PUBLIC_NETWORK_ENDPOINT` environment variable.
- Use `ipaddr:port` format for `seed_nodes`.
- Improve logging.


src/bin/stegosd/main.rs
```diff
@@ -369,6 +389,32 @@ fn run() -> Result<(), Error> {
...
+                    SocketAddr::from_str(&uri)
+                        .map(|_| ())
+                        .map_err(|e| format!("{}", e))
+                })
+                .takes_value(true),
+        )
+        .arg(
+            Arg::with_name("public-network-endpoint")
+                .long("public-network-endpoint")
+                .env("STEGOS_PUBLIC_NETWORK_ENDPOINT")
```
rtsisyk: I don't like this long name. Any better ideas? What is about `STEGOS_NETWORK_LISTEN`, `STEGOS_NETWORK_PUBLISH`?

network/src/libp2p_network/mod.rs
```diff
@@ -180,10 +183,11 @@ fn new_service(
         Swarm::new(transport, behaviour, peer_id)
     };
 
-    let mut bind_ip = Multiaddr::from(Protocol::Ip4(config.bind_ip.clone().parse()?));
-    bind_ip.push(Protocol::Tcp(config.bind_port));
-
-    Swarm::listen_on(&mut swarm, bind_ip).unwrap();
+    if config.endpoint != "" {
```
rtsisyk: Please note that I don't start listener if network.endpoint is empty. I tested on my local machine and it works.

src/bin/stegosd/main.rs
vldm: why you need public world in variable name?
LISTEN_ADDR should be ok as well.

network/src/libp2p_network/mod.rs
vldm: Maybe we should add log entry, with info that user has lack of some features.

network/src/ncp/behavior.rs
```diff
@@ -88,25 +90,32 @@ pub struct Ncp<TSubstream> {
...
-            .filter_map(|a| match a.parse() {
-                Ok(addr) => Some(addr),
-                Err(_) => None,
-            })
+            .map(|a| socket_to_multi_addr(&SocketAddr::from_str(a).expect("Invalid seed_nodes")))
             .collect();
+
+        let public_endpoint = if config.public_endpoint != "" {
+            let endpoint =
+                SocketAddr::from_str(&config.public_endpoint).expect("Invalid public_endpoint");
```
vldm: Would `SocketAddr::from_str` support hostnames?

rtsisyk: I don't think so. DNS support is a different story.

src/bin/stegosd/main.rs
rtsisyk: Discussed.

PR #1248, opened 2019-08-13
Persist UTXO history in blockchain.

vldm:


blockchain/src/blockchain.rs
```diff
@@ -152,7 +152,12 @@ pub struct OutputRecovery {
     pub timestamp: Timestamp,
 }
 
-pub type AccountRecoveryState = Vec<OutputRecovery>;
+#[derive(Eq, PartialEq, Debug, Default, Clone)]
+pub struct AccountRecoveryState {
+    pub removed: HashMap<Hash, Output>,
+    pub commited: Vec<OutputRecovery>,
+    pub prepared: Vec<OutputRecovery>,
```
rtsisyk: This structure is inconsistent - please either use HashMap<Hash, OutputRecovery> for all three members or Vec<OutputRecovery>. Probably the first option is better. 

Please also don't forget to remove `is_final` flag from `OutputRecovery`.

blockchain/src/blockchain.rs
```diff
@@ -381,11 +386,18 @@ impl Blockchain {
...
-        let mut accounts_state: Vec<AccountRecoveryState> =
-            vec![Default::default(); accounts.len()];
-        let mut epoch: u64 = 0;
+        account_skey: &SecretKey,
+        account_pkey: &PublicKey,
+        mut epoch: u64,
+        unspent: HashMap<Hash, Output>,
+    ) -> Result<AccountRecoveryState, BlockchainError> {
+        let mut accounts_state: AccountRecoveryState = AccountRecoveryState {
+            removed: unspent
```
rtsisyk: Why don't detect spent inputs during scanning the blockchain? The blockchain is scanned in any case. Check `block.inputs` during scanning to get rid of `unspent` argument.

node/src/api.rs
```diff
@@ -44,6 +44,10 @@ pub enum NodeRequest {
         account_skey: scc::SecretKey,
         /// Account Public Key.
         account_pkey: scc::PublicKey,
+        /// Map of unspent utxos, that account already know.
+        unspent_utxos: HashMap<Hash, Output>,
```
rtsisyk: Is not needed, I explain above why.

wallet/src/lib.rs
```diff
@@ -1120,22 +1196,19 @@ impl Future for UnsealedAccountService {
...
-                                    .payments
-                                    .iter()
-                                    .map(|(h, v)| v.to_info(self.pending_payments.get(h)))
-                                    .collect();
-                                let stakes: Vec<StakeInfo> = self
-                                    .stakes
-                                    .values()
-                                    .map(|value| value.to_info(epoch))
-                                    .collect();
+                                let mut public_payments = Vec::new();
```
rtsisyk: I think we can break output format here and return Vec<OutputInfo> instead.

wallet/src/storage.rs
```diff
@@ -41,8 +42,14 @@ use stegos_node::TransactionStatus;
 use stegos_serialization::traits::ProtoConvert;
 use tempdir::TempDir;
 
-const LEN_INDEX: [u8; 1] = [0; 1];
-const TIME_INDEX: [u8; 2] = [0; 2];
+// colon families.
+const ACCOUNT_LOG: &'static str = "account_log";
+const UNSPENT_UTXO: &'static str = "unspent_utxo";
+const META_CF: &'static str = "meta_cf";
+const COLON_FAMILIES: &[&'static str] = &[ACCOUNT_LOG, UNSPENT_UTXO, META_CF];
```
rtsisyk: Let's make column family names a.k.a table names more user friendly:
"account_log" => "history"
"unspent_utxo" => "unspent" (BTW, unspent UTXO is tautology!)
"meta_cf" => "meta"

wallet/src/storage.rs
```diff
@@ -41,8 +42,14 @@ use stegos_node::TransactionStatus;
...
+
+// indexes in meta cf
```
rtsisyk: Do you mean "keys"? Probably lowercase is better too, but it as arguable.

wallet/src/storage.rs
```diff
@@ -57,14 +64,12 @@ pub struct AccountLog {
...
-    /// Len of account log.
-    len: u64,
-    /// last known system time.
-    last_time: Timestamp,
 
     //
     // Indexes
     //
+    /// Index of epoch UTXOS, that is not final.
+    utxos: HashMap<Hash, UnspentOutput>,
```
rtsisyk: I stopped to understand what is the difference between `UnsealedAccountService` and `AccountLog` a.k.a. `AccountDB`:

AccountLog:

```
   /// Index of epoch UTXOS, that is not final.
    utxos: HashMap<Hash, UnspentOutput>,
    /// Index of UTXOS that known to be change.
    known_changes: HashSet<Hash>,
    /// Index of all created transactions by this wallet.
    utxos_list: HashMap<Hash, Timestamp>,
    /// Index of all created transactions by this wallet.
    created_txs: HashMap<Hash, Timestamp>,
    /// Index of all transactions that wasn't rejected or committed.
    pending_txs: HashSet<Hash>,
    /// Transactions that was created in current epoch.
    epoch_transactions: HashSet<Hash>,
```

AccountService:

```
    /// Current Epoch.
    epoch: u64,
    /// Time of last macro block.
    last_macro_block_timestamp: Timestamp,
    // Store time in Instant, to be more compatible with tokio-timer.
    /// List of pending utxos.
    pending_payments: HashMap<Hash, PendingOutput>,
```

Probably this abstraction doesn't work anymore. Similar approach failed in `struct Blockchain` and I had to merge `struct Storage` into `struct Blockchain`... What do you think about that?

PR #1250, opened 2019-08-13
Dm harden snowball

dbmcclain:
initial commit
code good, but needs to adjust test harness to account for new fee structure

### DIFF
--- a/wallet/src/lib.rs
+++ b/wallet/src/lib.rs
@@ -521,7 +521,9 @@ impl UnsealedAccountService {
             locked_timestamp,
             snowball::MAX_UTXOS,
         )?;
-        assert!(inputs.len() <= snowball::MAX_UTXOS);
+        // NOTE from DM: There is no particular limit to the numnber of TXINS imposed by Snowball.
+        // Only the number of TXOUTS per participant is limited and fixed to MAX_SHARING_TXOUTS
+        // assert!(inputs.len() <= snowball::MAX_SHARING_TXOUTS);
 
         let time = clock::now();
         for (input, _) in &inputs {
--- a/wallet/src/snowball/message.rs
+++ b/wallet/src/snowball/message.rs
@@ -42,6 +42,7 @@ pub(crate) enum SnowballPayload {
     SharedKeying {
         pkey: PublicKey,
         ksig: Pt,
+        fee: i64,
     },
     Commitment {
         cmt: Hash,
@@ -50,7 +51,6 @@ pub(crate) enum SnowballPayload {
     CloakedVals {
         matrix: DcMatrix,
         gamma_sum: Fr,
-        fee_sum: Fr,
         cloaks: HashMap<ParticipantID, Hash>,
     },
     Signature {
@@ -74,10 +74,10 @@ pub(crate) struct SnowballMessage {
 impl fmt::Display for SnowballPayload {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            SnowballPayload::SharedKeying { pkey, ksig } => write!(
+            SnowballPayload::SharedKeying { pkey, ksig, fee } => write!(
                 f,
-                "SnowballPayload::SharedKeying( pkey: {:?}, ksig: {:?})",
-                pkey, ksig
+                "SnowballPayload::SharedKeying( pkey: {:?}, ksig: {:?}, fee: {:?})",
+                pkey, ksig, fee
             ),
             SnowballPayload::Commitment { cmt, .. } => {
                 write!(f, "SnowballPayload::Commitment( cmt: {})", cmt)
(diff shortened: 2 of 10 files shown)


wallet/src/lib.rs
```diff
@@ -69,6 +69,9 @@ const RESEND_TX_INTERVAL: Duration = Duration::from_secs(2 * 60);
 const PENDING_UTXO_TIME: Duration = Duration::from_secs(5 * 60);
 const CHECK_PENDING_UTXO: Duration = Duration::from_secs(10);
 
+pub const MAX_SHARING_TXOUTS: usize = 5; // max nbr of txout UTXO permitted
+pub const FEE_PER_TXOUT: i64 = 1_000; // 0.001 STG = fee expected per txout UTXO shared
```
rtsisyk: The fee value is not fixed and provided by clients. There is no such thing as `FEE_PER_TXOUT` on the server. API clients can choose any arbitrary fee, greater than node.min_payment_fee in config.

wallet/src/snowball/message.rs
```diff
@@ -95,10 +95,10 @@ impl fmt::Display for SnowballPayload {
 impl fmt::Debug for SnowballPayload {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            SnowballPayload::SharedKeying { pkey, ksig } => write!(
+            SnowballPayload::SharedKeying { pkey, ksig, fee } => write!(
                 f,
-                "VsPayload::SharedKeying( pkey: {:?}, ksig: {:?})",
-                pkey, ksig
+                "VsPayload::SharedKeying( pkey: {:?}, ksig: {:?}, fee: {:?})",
```
rtsisyk: Please also remove "Vs" prefix.

wallet/src/snowball/mod.rs
```diff
@@ -72,6 +72,9 @@
 // ========================================================================
 
 #![allow(non_snake_case)]
+#![allow(unused)]
```
rtsisyk: Please don't add `#![allow(unused)]` - all unused code should be removed instead.

wallet/src/snowball/mod.rs
```diff
@@ -215,6 +216,18 @@ pub struct SnowballOutput {
     pub is_leader: bool,
 }
 
+#[derive(Debug, Clone)]
+pub enum SnowballTestPlan {
```
rtsisyk: I wrote in #1253 what I don't understand the idea of this struct. Please don't mix all patches into one.

wallet/src/snowball/mod.rs
```diff
@@ -383,7 +394,21 @@ impl Snowball {
         my_fee: i64,
     ) -> Snowball {
         // check the maximal number of UTXOs.
-        assert!(my_txouts.len() <= MAX_UTXOS);
+        assert!(my_txouts.len() <= MAX_SHARING_TXOUTS);
+        assert!(my_fee == (MAX_SHARING_TXOUTS as i64) * FEE_PER_TXOUT);
+
+        // fill out proposed UTXO's with dummies if fewer than
+        // consistent max number
+        let mut my_filled_txouts = my_txouts.clone();
```
rtsisyk: Why don't add `mut` to `my_txouts` instead?

wallet/src/snowball/mod.rs
dbmcclain: needed for instrumentation set_test_plan(), which nobody yet calls

wallet/src/snowball/mod.rs
dbmcclain: I did not mix, I based off of instrumentation PR

wallet/src/snowball/mod.rs
dbmcclain: fine by me...

wallet/src/lib.rs
dbmcclain: I won't dispute you... but why would anyone pay more than the minimum required fee?

wallet/src/snowball/message.rs
dbmcclain: I put VsPayload there because we will shortly also have a QsPayload for QueryShuffle. And BTW... you debug info messages don't tell you which shuffler the messages come from. There will be two possible shufflers running - Snowball and QueryShuffle. I added a prefix in the QueryShuffle debug messages (QS).

wallet/src/snowball/mod.rs
dbmcclain: The idea of this struct is to convey test-actions to a running session on a wallet node. You can tell it to fail to communicate with some or all other participants to force asymmetric or symmetric communication failures during testing.

dbmcclain:
I don't dispute anything about the Fees, but somebody must make a command decision here about how to arrive at fees for SuperTransactions and SuperQueries. I don't care how you choose to do this, but I need guidance.

Also need to decide about minimum fee spent, even if failure to make it into a final transaction or query. Otherwise, we are open to zero-cost flooding attacks.

dbmcclain:
Code conforms (mostly) to Roman's desires. Fees are now permitted to vary above some minimum acceptable level. Final super-transaction has sum of fees from remaining participants who made it into the supertransaction.

(Entire code base is predicated on hardcoded PAYMENT_FEE constant. Not yet using NodeConfig::min_payment_fee.)

rtsisyk:
> (Entire code base is predicated on hardcoded PAYMENT_FEE constant. Not yet using NodeConfig::min_payment_fee.)

That's not true. Only tests and CLI use PAYMENT_FEE. Node checks NodeConfig::min_payment_fee.

dbmcclain:
There currently exists no way for Snowball to obtain the node configuration, to obtain the minimum fee. Some additional argument must be required?

rtsisyk:
I'm moving this patch to Mainnet Beta 3. We want to focus on Windows port in Mainnet Beta 2.

rtsisyk:
I can't remember which problem did we want to fix in this patch?

rtsisyk:
Not actual anymore.

PR #1256, opened 2019-08-15
Enable backtrace and coredump by default

rtsisyk:


src/bin/stegosd/main.rs
```diff
@@ -273,7 +294,7 @@ fn load_configuration(args: &ArgMatches<'_>) -> Result<config::Config, Error> {
     }
 
     // Disable [chain] and [node] sections for mainnet and testnet.
-    let is_prod = cfg.general.chain == "mainnet" || cfg.general.chain == "testnet";
+    let is_prod = cfg.general.chain == "mainnet";
```
echupriyanov: What happened to `testnet`?

rtsisyk: I decided to allow everthing on testnet.

PR #1274, opened 2019-08-19
implement fixed size UTXO binary transfer across DiceMix

dbmcclain:


dbmcclain:
resync with dev branch

blockchain/protos/blockchain.proto
```diff
@@ -16,7 +16,7 @@ message PaymentOutput {
     stegos.crypto.Pt cloaking_hint = 2;
     stegos.crypto.BulletProof proof = 3;
     stegos.crypto.EncryptedPayload payload = 4;
-    uint64 locked_timestamp = 5;
+    repeated byte locked_timestamp = 5;
```
dbmcclain: this was in error - I will change back and repush

rtsisyk:
I'm moving this patch to Mainnet Beta 3. We want to focus on Windows port in Mainnet Beta 2.

blockchain/src/output.rs
```diff
@@ -121,6 +123,80 @@ pub struct PaymentOutput {
...
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut ans = self.recipient.to_bytes().to_vec();
+        ans.append(&mut self.cloaking_hint.to_bytes().to_vec());
+        ans.append(&mut self.proof.to_bytes().to_vec());
+        ans.append(&mut self.payload.to_bytes().to_vec());
+        let mut ts: u64 = match self.locked_timestamp {
+            None => 0u64,
+            Some(tsv) => tsv.into(),
+        };
+        for _ in 0..8 {
```
vldm: Can we use [byteorder](https://crates.io/crates/byteorder) crate for serializing simple types?
This code is not portable across platform that use different endianness

vldm: Sorry, this code is probably portable, but still it would be good to use byteorder, because it's widespread used in rust ecosystem

rtsisyk:
See https://github.com/stegos/stegos/pull/1323

rtsisyk:
Replaced by  #1323

PR #1279, opened 2019-08-20
Release both stripped and unstripped versions of binaries

rtsisyk:


rtsisyk:
> Semicolons are not necessary at the EOL inside the for loops

GitLab CI concatenates this multi-line expression into one line. Initially I had no semicolons in this expression, but it failed.

PR #1303, opened 2019-08-29
Fast synchronization protocol

rtsisyk:


rtsisyk:
Almost finished - moving to the next release.

Issue #1296, opened 2019-08-27
Node crashes if Snowball payment attempted during account recovery


echupriyanov:
If one tries to make Snowball payment immediately after unsealing the account, and account recovery is still in progress, node is crashing.

```
thread 'tokio-runtime-worker-11' panicked at 'assertion failed: self.snowball.is_none()', wallet/src/lib.rs:1258:25
stack backtrace:
   0:        0x101807452 - backtrace::backtrace::libunwind::trace::he2bd1d3c74d283d5
                               at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/libunwind.rs:88
   1:        0x101807452 - backtrace::backtrace::trace_unsynchronized::hc14993df006ef640
                               at /Users/vsts/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.34/src/backtrace/mod.rs:66
   2:        0x101807452 - std::sys_common::backtrace::_print::h926f1851a302f85a
                               at src/libstd/sys_common/backtrace.rs:47
   3:        0x101807452 - std::sys_common::backtrace::print::h40eeee4b52fc5d77
                               at src/libstd/sys_common/backtrace.rs:36
   4:        0x101807452 - std::panicking::default_hook::{{closure}}::h8b343d26c9df99a4
                               at src/libstd/panicking.rs:200
   5:        0x10180711d - std::panicking::default_hook::hf23341654d96249a
                               at src/libstd/panicking.rs:214
   6:        0x101807c30 - std::panicking::rust_panic_with_hook::h9defcda189ad7cc1
                               at src/libstd/panicking.rs:477
   7:        0x100e3e195 - std::panicking::begin_panic::h49a28e349058831c
                               at /rustc/d3f8a0b5dfddfe443d9db1f1da18348dbceb0e47/src/libstd/panicking.rs:411
   8:        0x100da5478 - <stegos_wallet::UnsealedAccountService as futures::future::Future>::poll::hfa9d3b82958407a8
                               at wallet/src/lib.rs:1258
   9:        0x100da613d - <stegos_wallet::AccountService as futures::future::Future>::poll::h0e3ec6350bbf2db0
                               at wallet/src/lib.rs:1471
  10:        0x101195bdf - <alloc::boxed::Box<F> as futures::future::Future>::poll::h3f3cf20c0a046d58
[...]

PR #1304, opened 2019-08-29
Block api on unseal.

vldm:
Closes: #1296


wallet/src/lib.rs
```diff
@@ -1001,6 +1006,63 @@ impl Future for UnsealedAccountService {
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         // Recovery information from node.
+        if !self.recovered {
```
rtsisyk: I can't say that I like this huge indented block.

vldm: I can split logic into separate method

PR #1311, opened 2019-09-06
patch around possibly missing rustfmt

dbmcclain:
nightly ARM64 does not have
does not affect production build


echupriyanov:
There is no `rustfmt` in _all_ nightly Rust builds for ARM

PR #1313, opened 2019-09-06
Implement SCC subkeying

dbmcclain:
fix erroneous use of Scalar::from_bits() to use Scalar::from_bytes_mod_order()

implement SCC subkeying for single chain derivation - simple u32 incremental
value applied to secret and public master keys


crypto/src/scc/mod.rs
```diff
@@ -850,6 +851,61 @@ pub fn make_random_keys() -> (SecretKey, PublicKey) {
...
+//   1. The pkey seed and the increment hash together to produce zero offset.
+//     This would produce no change and hence duplicate keying. Or,
+//   2. The increment would lead to a secret key of zero in field Fr, or,
+//   3. The increment would lead to a public key as the point at infinity,
+//
+//  Conditions 2 & 3 are actually the same, but approached from two different
+//  directions.
+//
+
+pub fn make_secret_subkey(skey: &SecretKey, index: u32) -> Result<SecretKey, CryptoError> {
```
rtsisyk: Let's convert it to a method of SecretKey.

crypto/src/scc/mod.rs
```diff
@@ -298,7 +298,7 @@ impl MulAssign<Fr> for Fr {
 
 impl From<Hash> for Fr {
     fn from(h: Hash) -> Fr {
-        Fr::from(Scalar::from_bits(h.bits()))
+        Fr::from(Scalar::from_bytes_mod_order(h.bits()))
```
rtsisyk: Please push a separate pull request. I will regenerate genesis...

crypto/src/scc/mod.rs
dbmcclain: Done - convert to trait method for both SecretKey and PublicKey

crypto/src/scc/mod.rs
dbmcclain: Genesis keying regenerated here

Issue #1281, opened 2019-08-20
Rewrite CLI using `clap`


rtsisyk:
Some CLI commands like `pay` are completely unusable with flags. Remove ad-hoc parse and use `clap` instead.

PR #1316, opened 2019-09-09
Implement Bitcoin-like public addresses - WIP

rtsisyk:
Needed for exchange integration.

### DIFF
--- a/crypto/src/scc/mod.rs
+++ b/crypto/src/scc/mod.rs
@@ -878,6 +878,14 @@ impl PartialOrd for PublicKey {
     }
 }
 
+impl std::hash::Hash for PublicKey {
+    // Needed to use std::collections::HashMap.
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        std::hash::Hash::hash("PublicKey", state);
+        std::hash::Hash::hash(&self.0.to_bytes(), state);
+    }
+}
+
 // -----------------------------------------------------------------------
 // Key Generation & Checking
 
--- a/src/bin/stegos/console.rs
+++ b/src/bin/stegos/console.rs
@@ -28,7 +28,7 @@ use rpassword::prompt_password_stdout;
 use rustyline as rl;
 use serde::ser::Serialize;
 use std::fmt;
-use std::io::stdin;
+use std::io::{stdin, stdout, Write};
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::{Arc, Mutex};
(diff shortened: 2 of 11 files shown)


crypto/src/scc/mod.rs
```diff
@@ -878,6 +878,14 @@ impl PartialOrd for PublicKey {
     }
 }
 
+impl std::hash::Hash for PublicKey {
```
dbmcclain: Do we know what algorithm std::hash::Hasher uses? The other day Vladimir suggested that it uses SipHash. But online docs indicate that this is Deprecated.

src/bin/stegos/console.rs
```diff
@@ -768,6 +765,12 @@ impl ConsoleService {
         } else if msg == "restake" {
```
dbmcclain: Personally, the current handling of commands seems too rigid. Why not allow case insensitive unique prefix abbreviations, get away from positional arguments? I found e.g., that /snowball had to be placed "after" other args, but had to be placed before /fee (or do I have it backwards...?)

wallet/src/storage.rs
```diff
@@ -508,14 +509,47 @@ impl AccountDatabase {
...
+        batch.put_cf(
+            meta_cf,
+            LAST_PUBLIC_ADDRESS_ID_KEY,
+            &Self::bytes_from_u32(last_public_address_id),
+        )?;
+        self.database.write(batch)?;
+        Ok(())
+    }
+
     /// Convert timestamp to bytearray.
```
dbmcclain: You seem to be implicitly expecting that all byte strings will be Big Endian. This cannot hold for some crypto items (unless you explicitly rewrite conversions for underlying machinery which expects little-endian order).

But I do find it excessive that Rust demands such converters be manually written. Depending on architecture, e.g., a u64 is its own byte string already of some endianness.

There have been some long-standing conventions, e.g., network byte order which IIRC is big-endian.

And even with a selected convention, what order does a hash value present? And for that matter, what order does any finite field value represent? (circular modular arithmetic, like a clock, does 4 o'clock come before or after 7 o'clock?)

crypto/src/scc/mod.rs
rtsisyk: I have no idea. I implemented this Trait just to use HashMap.

src/bin/stegos/console.rs
rtsisyk: Just because we have no time. 
See https://github.com/stegos/stegos/issues/1281

wallet/src/storage.rs
rtsisyk: I don't understand where I expect that all byte strings are Big Endian. Please elaborate.

PR #1326, opened 2019-09-18
Replace Base58 to Bech32 implementation

vldm:


crypto/src/lib.rs
```diff
@@ -121,16 +120,64 @@ impl From<hex::FromHexError> for CryptoError {
...
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+static STATE: AtomicUsize = AtomicUsize::new(UNINITIALIZED);
+const UNINITIALIZED: usize = 0;
+const INITIALIZING: usize = 1;
+const INITIALIZED: usize = 2;
+
+fn chain_to_prefix(network: &str) -> &'static str {
+    match network {
```
rtsisyk: I propose to move this logic to main.rs

vldm: Its reused in bootstrap/node_test etc

PR #1335, opened 2019-09-25
Faster load from disk

vldm:


blockchain/src/blockchain.rs
```diff
@@ -258,7 +282,12 @@ impl Blockchain {
         //
         // Storage.
         //
-        let database = rocksdb::DB::open_default(chain_dir)?;
+
+        let mut opts = rocksdb::Options::default();
```
rtsisyk: Please try to propagate I/O error to the upper level. We should avoid using unwrap() for I/O errors. Disk error are not rare, panic!() is bad solution for them.

blockchain/src/blockchain.rs
```diff
@@ -330,34 +357,124 @@ impl Blockchain {
...
-        let genesis_hash = Hash::digest(&genesis);
+        let cf_block_by_hash = self.database.cf_handle(BLOCK_BY_HASH).unwrap();
+        let cf_output_by_hash = self.database.cf_handle(OUTPUT_BY_HASH).unwrap();
+        let cf_escrow = self.database.cf_handle(ESCROW).unwrap();
+        let cf_meta = self.database.cf_handle(META).unwrap();
+
+        // TODO: Remove box in type;
+        let blocks: Box<dyn Iterator<Item = Block>> =
+            if let Some(_) = self.database.get_cf(cf_meta, EPOCH.as_bytes())? {
+                macro_rules! recover_meta {
```
rtsisyk: This large indentation block doesn't look nice. Try to refactor this code by splitting "recovery" and "bootstrap" cases.

blockchain/src/blockchain.rs
```diff
@@ -330,34 +357,124 @@ impl Blockchain {
...
+                    .is_none());
+                // balance already set, that's why dont assert
+                let _ = self.balance.insert(lsn, (), recover_meta!(BALANCE));
+                recover_map!(cf_block_by_hash, self.block_by_hash, lsn);
+                recover_map!(cf_output_by_hash, self.output_by_hash, lsn);
+                let mut escrow = EscrowMap::new();
+                recover_map!(cf_escrow, escrow, lsn);
+                self.escrow.escrow = escrow;
+
+                let mut new_lsn = lsn;
```
rtsisyk: Please add some comments here.

blockchain/src/blockchain.rs
```diff
@@ -330,34 +357,124 @@ impl Blockchain {
...
+                        genesis_hash,
+                        self.last_block_hash(),
+                    )
+                    .into());
+                }
+                Box::new(blocks)
+            };
         // Recover remaining blocks.
         for block in blocks {
             self.recover_block(block, timestamp, force_check)?;
```
rtsisyk: I see that we still iterate over the blocks. The idea was to don't read entire blockchain on the start. This problem is a showstopper.

blockchain/src/blockchain.rs
```diff
@@ -606,8 +723,16 @@ impl Blockchain {
...
-        self.epoch_infos.get(&epoch)
+    pub fn epoch_info(&self, epoch: u64) -> Option<EpochInfo> {
+        let cf_epoch_infos = self.database.cf_handle(EPOCH_INFOS).unwrap();
+        let epoch_info = self
+            .database
+            .get_cf(
+                cf_epoch_infos,
+                &Self::block_key(LSN(epoch, MACRO_BLOCK_OFFSET)),
+            )
+            .ok()?;
```
rtsisyk: Please don't try and hide I/O errors in such way. If function can fail - it should return Result<Something, BlockchainError>.

blockchain/src/blockchain.rs
```diff
@@ -1241,28 +1365,50 @@ impl Blockchain {
                 .set(*stake);
         }
 
+        let cf_block_by_hash = self.database.cf_handle(BLOCK_BY_HASH).unwrap();
```
rtsisyk: I'm not sure about unwrap() here. Probably this error should be propagated to the upper level, doesn't it?

blockchain/src/blockchain.rs
```diff
@@ -1758,6 +1911,50 @@ impl Blockchain {
...
+    where
+        V: ProtoConvert + std::fmt::Debug,
+    {
+        let value = value.into_buffer()?;
+        batch.put_cf(meta_cf, key.as_bytes(), &value)?;
+        Ok(())
+    }
+
+    /// Undolog is actualy a patchset, so just apply it to the block.
+    // TODO: compress patch before merging.
```
rtsisyk: What should be done here? Why it hasn't done?

blockchain/src/mvcc.rs
```diff
@@ -279,6 +281,31 @@ where
...
+    pub fn reset(&mut self) {
+        self.map.clear();
+        self.undo.clear();
+        self.checkpoint_lsn = Default::default();
+    }
+
+    // Convert UndoLog into Diff.
+    fn reverse_patch(&self, undo: Vec<UndoRecord<K, V, LSN>>) -> BTreeMap<K, Option<V>> {
+        let mut result = BTreeMap::new();
+        for record in undo {
```
rtsisyk: Are you sure that you are iterating in reverse order? it would be nice to test it somewhere.

blockchain/src/protos.rs
```diff
@@ -43,10 +43,365 @@ pub enum ProtoError {
 }
 
 // link protobuf dependencies
+use crate::awards::Awards;
+use std::collections::BTreeMap;
 use stegos_crypto::protos::*;
 use stegos_crypto::CryptoError;
 include!(concat!(env!("OUT_DIR"), "/protos/mod.rs"));
 
+impl ProtoConvert for LSN {
```
rtsisyk: Protobuf needs basic round-trip tests because you can easily forget to serialize some field and nobody will note.

blockchain/src/blockchain.rs
```diff
@@ -330,34 +357,124 @@ impl Blockchain {
...
+                    return Err(BlockchainError::IncompatibleGenesis(
+                        genesis_hash,
+                        self.last_block_hash(),
+                    )
+                    .into());
+                }
+                Box::new(blocks)
+            };
         // Recover remaining blocks.
         for block in blocks {
```
rtsisyk: Please remove `recover_block()`. A part related MicroBlock should go to the "recovery" case. A part related to MacroBLock should go to "bootstrap" case. Otherwise it's not clear that constructor doesn't try to read the entire blockchain from the disk.

blockchain/src/blockchain.rs
vldm: If `cf_handle` cannot be found - then database was created without cf.

blockchain/src/blockchain.rs
vldm: It's already done

blockchain/src/mvcc.rs
vldm: It's no differences in what order i iterate, i need only updated values of changed keys.

rtsisyk:
> This ensures that senders do not overwhelm receivers and keeps buffer usage low. However, depending on the protocol, there is a risk of deadlock, namely if both endpoints want to send data larger than the receivers window and they do not read before finishing their writes. Use this mode only if you are sure that this will never happen, i.e. if
>
>    Endpoints A and B never write at the same time, or
>   Endpoints A and B write at most n frames concurrently such that the sum of the frame lengths is less or equal to the available credit of A and B respectively.

> "A and B never write at the same time".

 This is a socket, not a pipe. Both sides can write at the same time. Requested conditions can't be archived.

Anyway, how does this flag help to fix this bug:

https://github.com/paritytech/yamux/blob/master/src/connection.rs#L458-L472:

```
                if stream.buffer.lock().len().map(move |n| n >= max_buffer_size).unwrap_or(true) {
                    error!("{}: {}: buffer of stream grows beyond limit", self.id, stream_id);
                    self.reset(stream_id)?;
                    return Ok(None)
                }
```

Here Yamix returns EOF instead of stoping reading this stream, But, wait a miniute. Houston, we have a problem here. Yamux can't read streams selectively because ... there is a real  TCP SOCK_STREAM under the hood. You can't read SOCK_STREAM selectively. Surprise! There is no way to control the flow of each specific stream. You need SOCK_DGRAM of raw packets to do that instead of SOCK_STREAM. Looks like this idea with multiple TCP-like streams with custom flow control in each stream is just broken by design.  Please correct me if I'm wrong.

PR #1338, opened 2019-09-26
Resolve network name in cli from node config.

vldm:


rtsisyk:
Moving to the next release, can't merge it in the last seconds.

PR #1350, opened 2019-09-30
Disable SSE for rust_crypto.

vldm:


rtsisyk:
We need a better solution rather than disabling entire SSE for everybody.

PR #1368, opened 2019-10-15
Fix "entered unreachable code: Expected Connecting state"

rtsisyk:


node/src/replication/peer.rs
```diff
@@ -160,7 +164,11 @@ impl Peer {
             Peer::Connecting {
                 peer_id, multiaddr, ..
             } => (peer_id, multiaddr),
-            _ => unreachable!("Expected Connecting state"),
+            _ => {
+                // Unexpected state - disconnect.
+                self.disconnected();
+                return;
```
vldm: you can use 

```
return self.disconnected();
```

rtsisyk: OK

PR #1373, opened 2019-10-17
Fix BitVec serialization and view_change proof sending mechanism.

vldm:


crypto/src/utils/mod.rs
```diff
@@ -228,6 +228,16 @@ where
     serializer.serialize_str(&string_vec)
 }
 
+/// Remove all trailing zeros from bitvec.
+pub fn trim_bitvec(vec: &mut BitVec) {
+    while let Some(last) = vec.pop() {
```
rtsisyk: while let Some(last) = vec.last() {
   vec.pop();
}

vldm: Where you find method last ?

blockchain/src/multisignature.rs
```diff
@@ -96,6 +97,7 @@ pub fn check_multi_signature(
 ) -> Result<(), MultisignatureError> {
     // Check for trailing bits in the bitmap.
     if multisigmap.len() > validators.len() {
+        warn!("multisigmap = {:?}", multisigmap);
```
rtsisyk: Please remove.

PR #1375, opened 2019-10-17
Use Cargo to build libgmp, libmpfr, libflint

rtsisyk:


ci-scripts/build.sh
```diff
@@ -292,33 +190,16 @@ install_rocksdb_mingw() {
 
 # Build dependencies on Linux
 install_libraries_linux() {
-    install_gmp_linux
-    install_mpfr_linux
-    install_flint_linux
+    echo
```
vldm: why not just removing this method?

rtsisyk: Lets do it later after merging RocksDB stuff.

PR #1376, opened 2019-10-17
Re-add transactions into mempool() during apply_macro_block()

rtsisyk:


node/src/lib.rs
```diff
@@ -1286,41 +1256,50 @@ impl NodeService {
...
-            let msg = RevertedMicroBlock {
-                block,
-                transaction_statuses,
-                pruned_outputs,
-                recovered_inputs,
-            };
-            notify_subscribers(&mut self.chain_subscribers, msg.into());
-            self.update_validation_status()
-        } else {
+        if self.chain.offset() == 0 {
```
vldm: missing early return on if block

PR #1384, opened 2019-10-24
Fix synchronized status after notebook sleep.

vldm:


node/src/lib.rs
```diff
@@ -2064,6 +2069,27 @@ impl Future for NodeService {
             error!("Error: {}", e);
         }
 
+        loop {
+            match self.check_sync.poll() {
+                Ok(Async::Ready(Some(_))) => {
+                    if !self.is_synchronized() {
+                        self.on_status_changed();
```
rtsisyk: This line will send notifications every 30 seconds.

vldm: Yes, if synchronized was not reached, do you propose to cache last time and status?

PR #1392, opened 2019-10-30
Stegos Chat

dbmcclain:
a work in progress

### DIFF
--- a/blockchain/src/chat.rs
+++ b/blockchain/src/chat.rs
@@ -238,7 +238,7 @@ pub enum OutgoingChatPayload {
     // payload was decrypted into a list of evicted members
     Evictions(Vec<PublicKey>),
     // payload for telling group about new member (pkey,chain) info
-    NewMembers(Vec<(PublicKey, Hash)>),
+    NewMembers(Vec<(PublicKey, Fr)>),
     // payload has been decrypted into plaintext of whatever content
     // leave room for 1 byte of SubMsgType info in encrypted payload
     PlainText(Vec<u8>),
@@ -247,9 +247,9 @@ pub enum OutgoingChatPayload {
 #[derive(Clone)]
 pub enum IncomingChatPayload {
     Evictions(Vec<PublicKey>),
-    NewMembers(Vec<(PublicKey, Hash)>),
+    NewMembers(Vec<(PublicKey, Fr)>),
     PlainText(Vec<u8>),
-    Rekeying(Hash),
+    Rekeying(Fr),
 }
 
 /// ChatMessageOutput canary  canary for the light nodes..
@@ -290,9 +290,9 @@ impl ChatMessageOutput {
     pub fn cloak_recipient(
         &mut self,
         owner_pkey: &PublicKey,
-        owner_chain: &Hash,
+        owner_chain: &Fr,
         r_owner: &Fr,
-        sender_chain: &Hash,
+        sender_chain: &Fr,
     ) {
         // key_hint = random_A * PKey_owner
         // cloaked_owner = key_hint * Chain_owner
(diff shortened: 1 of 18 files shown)


blockchain/src/chat.rs
```diff
@@ -0,0 +1,222 @@
...
+    // Cloaked pubkey of sender
+    pub sender: PublicKey,
+
+    /// Creation time of output.
+    pub created: Timestamp,
+
+    /// Message sequence ID
+    pub sequence: u64,
+    // Segmentation info msg #x of #n
+    pub msg_nbr: u32,
```
rtsisyk, line 89: I remember that we discussed that we don't want segmentation in the first version...

blockchain/src/chat.rs
```diff
@@ -0,0 +1,222 @@
...
+    aes_decrypt, aes_decrypt_with_rvalue, aes_encrypt, sign_hash, validate_sig, EncryptedPayload,
+    Fr, Pt, PublicKey, SchnorrSig, SecretKey,
+};
+use stegos_crypto::CryptoError;
+
+const CHAT_MESSAGE_LIFETIME: Duration = Duration::from_secs(30 * 86400);
+
+/// Chat Message UTXO.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct ChatMessageOutput {
```
rtsisyk, line 71: Why a new type of UTXO is needed? We inteionally increased the size of PaymentUTXO to 2k just to fit potential change messages!

blockchain/src/chat.rs
dbmcclain: Yes we did. But that doesn't mean we will hold off forever... Just making allowance for future growth here.

blockchain/src/chat.rs
dbmcclain: There are things needed to be shown (e.g., sender pkey, creation date, message ID, signature) at the outer level of the Chat UTXO that have to be seen by validators and group members in order to validate and decrypt these special UTXO. The original payment UTXO was never designed with Chat in mind.

blockchain/src/chat.rs
dbmcclain: There is one category of Chat messages that will produce benign fragmentation when you have more than 59 members in the group. It is helpful if all of the rekeying messages from one sender are identified with the same messsage ID, part M of N. Each fragment stands alone and does not need reassembly, but they are pieces of one interchange from the sender.

rtsisyk:
Any news here? Do we have anything for review?

dbmcclain:
To give a bit of background on the latest changes...

I have decided that, in order to avoid problems with reading out-of-order messages and stale messages, it is best if group owners initially set up a special chain code for use in all group rekeyings. 

That way, all new members will be informed of: owner pkey, owner chain, rekey chain, then list of all other group members showing public key, current chain of member, and epoch of last update seen by owner. This information is relayed through a private message from owner to new member.

Upon receipt of this information, the new member will construct a ChatSession struct, and add it to his Chat struct by calling subscribe_to_group(). 

Since the new member will not be in any other member's roster yet, the group owner must send a group message informing them all, of the new members public key and initial chain code. If we allow group owner to assign the first chain code for the new member, then we can avoid a rekeying storm from the new member.
 
The rekeying chain for the group never needs to change. If any member becomes evicted he will still be able to read all of the rekeying messages, but since at least one member - the most important one for crypto = group owner (who sent the eviction notice) - the primary member will not be sending a new chain code to the evicted member's public key. 

So even if all other members are still ignorant of the eviction, and they include the evicted member in their rekeyings, the evicted member will not know the new chain code of the group owner, to which all future group messages will be sent.

Hence, knowing the chain codes for all other members becomes useless since evicted member won't be able to find any group messages under the new rekeyings.

vldm:
@dbmcclain Lets, close this pullrequest for now and keep base branch for future development.

PR #1403, opened 2019-11-06
Fix service award bug.

vldm:
Add tests for service award state.
Improve test of bitvec, for bigger collections.


src/bin/stegosd/main.rs
```diff
@@ -525,6 +532,11 @@ fn run() -> Result<(), Error> {
                 .help("Force BP, BLS, VRF validation during recovery")
                 .long("force-check"),
         )
+        .arg(
+            Arg::with_name("force-load-chain")
+                .help("Force recovery using blocks saved on disk, rather than Snapshot.")
+                .long("force-load-chain"),
```
rtsisyk: May be just `--recover` ?

PR #1404, opened 2019-11-07
Collapse stake outputs into single during restake.

vldm:


node/src/lib.rs
```diff
@@ -674,30 +674,44 @@ impl NodeService {
...
+                "Adding output info to accumulator: amount={}, key={}",
+                amount,
+                account_pkey
             );
+            match &mut output_info {
+                None => {
+                    output_info = Some((*account_pkey, amount));
+                }
+                Some(o) => {
+                    assert_eq!(&o.0, account_pkey, "account key should be same");
```
rtsisyk: This assertion is not true. Nobody prevents node stakes with multiple account_pkey and the same network_pkey in our system. We need to add this limit first.

vldm: We already has limit that multiple network_keys can has only 1 account key.
https://github.com/stegos/stegos/blob/dev/blockchain/src/escrow.rs#L283-L284

PR #1408, opened 2019-11-11
Add metrics for consensus and macroblock.

vldm:


consensus/src/metrics.rs
```diff
@@ -46,6 +46,24 @@ lazy_static! {
     )
     .unwrap();
 
+    pub static ref PREVOTES_AMOUNT: IntGauge = register_int_gauge!(
+        "stegos_prevotes_amount",
```
rtsisyk: stegos_consensus_prevotes ?

PR #1409, opened 2019-11-11
Add canaries for the light nodes

rtsisyk:


blockchain/protos/blockchain.proto
```diff
@@ -13,10 +13,10 @@ message PaymentPayloadData {
 
 message PaymentOutput {
     stegos.crypto.PublicKey recipient = 1;
-    stegos.crypto.Pt cloaking_hint = 2;
-    stegos.crypto.BulletProof proof = 3;
-    stegos.crypto.EncryptedPayload payload = 4;
-    uint64 locked_timestamp = 5;
+    stegos.crypto.BulletProof proof = 2;
+    uint64 locked_timestamp = 3;
+    stegos.crypto.Pt ag = 4;
```
vldm: why rename to ag?

blockchain/src/output.rs
```diff
@@ -100,26 +101,20 @@ pub struct PaymentOutput {
...
-    /// Size is approx 137 Bytes =
-    ///     (R-val 65B, crypto-text 72B = (amount 8B, gamma 32B, delta 32B))
-    pub payload: EncryptedPayload,
-
     /// Timelock for output.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub locked_timestamp: Option<Timestamp>,
+
+    /// AES keying hint needed to decrypt payload.
+    pub ag: Pt,
```
vldm: Could you also add some test for keying hint?

blockchain/src/block.rs
```diff
@@ -185,17 +231,44 @@ impl MicroBlock {
...
+        let mut canary_hashes = Vec::with_capacity(2 * transactions.len());
+        for tx in transactions {
+            let mut hasher = Hasher::new();
+            tx.fullhash(&mut hasher);
+            transaction_hashes.push(hasher.result());
+            input_hashes.extend(tx.txins().iter().cloned());
+            output_hashes.extend(tx.txouts().iter().map(Hash::digest));
+            canary_hashes.extend(tx.txouts().iter().map(|o| Hash::digest(&o.canary())));
+        }
+        let transactions_range_hash = Merkle::root_hash_from_array(&transaction_hashes);
```
vldm: Assert that canary_len = outputs_len?

blockchain/src/validation.rs
```diff
@@ -539,6 +536,17 @@ impl Blockchain {
             .into());
         }
 
+        // Sic: the following fields can't be validated properly
+        // without processing the block itself:
```
vldm: It's okay, just need to write that validate_macro_block_header only check compatibility of block with current blockchain.

blockchain/src/output.rs
vldm: I see a test for cannry in next commits, so we probably can go without separate test for hint.

blockchain/protos/blockchain.proto
rtsisyk: No renaming here.  I inlined EncryptedPayload. Why this field is called `ag`? Probably because it "alpha * G". I can rename it...

blockchain/src/block.rs
rtsisyk: It was here! ))

blockchain/src/validation.rs
rtsisyk: I refactored this part.

PR #1417, opened 2019-11-15
Fix rust toolchain, to use new version of gcc.

vldm:


ci-scripts/cargo-config
```diff
@@ -1,19 +1,19 @@
...
 
 [target.x86_64-linux-android]
 ar = "x86_64-linux-android-ar"
-linker = "x86_64-linux-android-rustlinker"
+linker = "rustlinker"
 
 [target.i686-linux-android]
 ar = "i686-linux-android-ar"
-linker = "i686-linux-android-rustlinker"
+linker = "rustlinker"
```
rtsisyk: Please note that rustlinker uses its name to determine which CC to call. For example, x86_64-linux-android-rustlinker calls x86_64-linux-android-cc, arm-linux-androideabi-rustlinker calls arm-linux-androideabi-cc and so on.

rtsisyk (changes requested):

rtsisyk (approved):

PR #1423, opened 2019-11-19
Add a sharedlibrary + JNI bindings

rtsisyk:


vldm (approved):

rtsisyk:
I see that shared libraries are broken on Android. I will try to fix it.

PR #1426, opened 2019-11-20
Fr from Hash

dbmcclain:
change hash to Fr to utilize right shifing of hash value to range (2^251, 2^252)

        // With this change, 97% of all hashes will
        // produce an Fr value in the range (2^251, 2^252). 
        // Whereas, before, a modular wrapping would keep only
        // about 50% of all hashes in that range.
        //
        // Hence we significantly increase the probability of
        // having many mixing bits in Fr multiplications.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1189,7 +1189,7 @@ dependencies = [
 
 [[package]]
 name = "libp2p"
-version = "0.1.0"
+version = "1.0.0"
 dependencies = [
  "futures 0.1.29 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


rtsisyk:
This change probably breaks compatibility in dicemix. I see that Fr::from(Hash) is used by dc_slot_pad() which is part of dicemix. I'm not sure that we should break compatibility without a good reason.

rtsisyk (approved):

rtsisyk:
@dbmcclain , I approved this PR, but please ensure that it doesn't break compatibility on the binary level of snowball protocol.

vldm:
This is a breaking change, and it didn't fixes high priority bugs so i think we can close this task.

PR #1427, opened 2019-11-21
Add stake all command, and info about inputs limit.

vldm:
Closes: #169890572 (pivotal), #169393032 

TODO:
- Need tests.

### DIFF
--- a/wallet/src/error.rs
+++ b/wallet/src/error.rs
@@ -28,7 +28,7 @@ use stegos_crypto::scc;
 pub enum WalletError {
     #[fail(display = "Duplicate account: pkey={}", _0)]
     DuplicateAccount(scc::PublicKey),
-    #[fail(display = "Not enough money.")]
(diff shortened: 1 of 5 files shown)


wallet/src/error.rs
```diff
@@ -28,7 +28,7 @@ use stegos_crypto::scc;
 pub enum WalletError {
     #[fail(display = "Duplicate account: pkey={}", _0)]
     DuplicateAccount(scc::PublicKey),
-    #[fail(display = "Not enough money.")]
+    #[fail(display = "Not enough money. Or outputs limit is reached.")]
```
rtsisyk, line 31: I don't see how this new error message is used.

vldm: This is the same error variant, but we used it before incorrectly, we always have limit of inputs that can be created in TX. When this limit is reached account will report that there is no enough money, which is incorrect.

rtsisyk (approved):

PR #1430, opened 2019-11-22
Resend restake transactions, that stucks.

vldm:


rtsisyk (approved):
This fix is pretty simple!

PR #1434, opened 2019-11-25
Fix Snowball to avoid DiceMix panics

dbmcclain:
This patch provides error handling from DiceMix errors, instead of panicking. It doesn't solve the reason for why the errors are happening. But if an error arises in DiceMix, the attempt at forming a provisional super-transaction is halted, a log message is printed, and the session is restarted for that node.

In all likelihood this will eventually lead to the node becoming orphaned from the group, and he will fail this Snowball attempt. He will then have to renegotiate between wallet and facilitator to try again in another Snowball session.

I can only guess that the reason for the DiceMix errors are that the node failed to see some participants and now receives cloaked data from others, which includes cloaking for the nodes he doesn't know about. And since their cloaking includes the participants not known to the node, he can't successfully uncloak the data. He then has no choice but to give up and try again for another session where he can hear all the participants.

### DIFF
--- a/crypto/src/dicemix/mod.rs
+++ b/crypto/src/dicemix/mod.rs
@@ -31,12 +31,14 @@ use crate::hash::{Hash, Hashable, Hasher};
 use crate::pbc;
 use crate::scc;
 use crate::utils::u8v_to_hexstr;
+use crate::CryptoError;
 use rayon::prelude::*;
 use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::ffi::CString;
 use std::fmt;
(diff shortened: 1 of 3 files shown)


rtsisyk (approved):
Great! But please propagate Dicemix to the upper level. Current implementation just suppresses this error:

```
       ) {
            Err(_) => {
                swarn!(self, "DiceMix Error");
                return self.start();
            }
            Ok(ms) => ms,
        };
```

dbmcclain:
Roman, I don't understand what you mean by "propagate to upper level". This suppresses the error, reports it in the log, and restarts the protocol for a new session at that node. There is no upper level. ???

PR #1445, opened 2019-12-05
Read only accounts 

vldm:
This design is currently pending. 
But still creating a pullrequest for review.

### DIFF
--- a/blockchain/src/blockchain.rs
+++ b/blockchain/src/blockchain.rs
@@ -729,6 +729,57 @@ impl Blockchain {
             .map(|r| r.output))
     }
 
+    /// Get accounts public utxos, collection.
+    pub fn get_public_unspent(
+        &self,
+        account_pkey: scc::PublicKey,
+    ) -> impl Iterator<Item = Result<PublicPaymentOutput, StorageError>> + '_ {
+        self.output_by_hash.iter().filter_map(move |(_hash, key)| {
+            match key {
+                OutputKey::MacroBlock { epoch, output_id } => {
+                    let block = match self.macro_block(*epoch) {
+                        Ok(b) => b,
+                        Err(e) => return Some(Err(e)),
+                    };
+                    assert_eq!(block.header.epoch, *epoch);
+                    match block.outputs.get(*output_id as usize) {
+                        Some(Output::PublicPaymentOutput(output))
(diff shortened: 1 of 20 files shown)


blockchain/src/blockchain.rs
```diff
@@ -729,6 +729,57 @@ impl Blockchain {
             .map(|r| r.output))
     }
 
+    /// Get accounts public utxos, collection.
+    pub fn get_public_unspent(
```
rtsisyk, line 733: This implementation is very inefficient. I don't think we can scan entire outputs_by_hash list one by one. It won't work on the real mainnet. It would be better to save a hint or even an entire public key for each PublicPaymentOutput in outputs_by_hash map.

wallet/src/api.rs
```diff
@@ -212,7 +228,7 @@ pub enum AccountRequest {
     },
     AccountInfo {},
     BalanceInfo {},
-    PublicAddressesInfo,
+    PublicAddressInfo,
```
rtsisyk, line 231: What is "Public Address". I thought that this feature will be completely removed.

wallet/src/api.rs
```diff
@@ -278,16 +298,28 @@ pub enum AccountResponse {
     Unsealed,
     #[serde(skip)]
     Disabled,
-    PublicAddressCreated {
+    ExportReadOnlyAccount {
+        public_address: scc::PublicKey,
+    },
+    PublicAddressInfo {
```
rtsisyk, line 304: Looks like AccountInfo works for ReadOnlyAccounts too. There is no need in this additional method.

wallet/src/api.rs
```diff
@@ -169,25 +168,42 @@ pub enum AccountRequest {
     Unseal {
         password: String,
     },
-    CreatePublicAddress,
+    ExportReadOnlyAccount,
```
rtsisyk, line 171: This method does the same as "AccountInfo".

rtsisyk (changes requested):
- Please remove "Implement api for work with raw transaction" from this PR.
- get_public_unspent() is super inefficient.
- Some AccountInfo/ExportPublicAddess/PublicAddress do the same job, let's keep the first one.

rtsisyk:
This feature is no longer needed.

PR #1447, opened 2019-12-06
Fix assertion panic when not all UTXOs make it across DiceMix sharing

dbmcclain:
Instead of assert(), which can panic, we detect that not all UTXO's came across and perform a session restart. Note that all participants will eventually detect a failed transaction balance condition, if they don't see their own UTXOs missing. So all participants will restart eventually.

### DIFF
--- a/wallet/src/error.rs
+++ b/wallet/src/error.rs
@@ -45,10 +45,7 @@ pub enum WalletError {
         _1, _0
     )]
     NoEnoughToStake(i64, i64),
-    #[fail(
-        display = "{} tokens available out of {}. Not enough to pay!",
-        _1, _0
-    )]
+    #[fail(display = "{} tokens available out of {}. Not enough to pay!", _1, _0)]
     NoEnoughToPay(i64, i64),
     #[fail(display = "{} tokens is not enough for a public payment!", _0)]
     NoEnoughToPayPublicly(i64),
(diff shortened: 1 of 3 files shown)


wallet/src/error.rs
```diff
@@ -45,10 +45,7 @@ pub enum WalletError {
         _1, _0
```
dbmcclain, line 45: I did not do this...perhaps from cargo fmt?

wallet/src/lib.rs
```diff
@@ -629,9 +629,11 @@ impl UnsealedAccountService {
     fn unstake(&mut self, amount: i64, payment_fee: i64) -> Result<TransactionInfo, Error> {
```
dbmcclain, line 629: I did not do this... again, perhaps from cargo fmt

wallet/src/snowball/mod.rs
```diff
@@ -1522,12 +1522,23 @@ impl Snowball {
         });
```
dbmcclain, line 1522: I did this

vldm (approved):

wallet/src/error.rs
vldm: Yes, sorry for that, Joel in his previous patch, forgot to apply rustfmt.

But thanks that you fix it!

PR #1448, opened 2019-12-06
Change crate that implement AES

vldm:


crypto/src/aont/mod.rs
```diff
@@ -25,7 +25,10 @@ use crate::hash::*;
 use crate::CryptoError;
```
dbmcclain, line 25: I don't believe that AONT encryption is being used anywhere

dbmcclain (approved):
looks okay to me

crypto/src/aont/mod.rs
vldm: + But in this pullrequest i tried to don't broke anything (even not used). We can remove it later.

PR #1451, opened 2019-12-17
chat members with epochs

dbmcclain:
Members in group membership rosters now keep PublicKey, Chain Hash, and Epoch (Timestamp) of last update. This is hoped to avoid issues with reading of stale rekeying messages. Rekeyings will only have an effect if the UTXO create time is greater than the member recorded epoch.


dbmcclain:
The function subscribe_to_group() has been removed. On initial join, the group owner should have told us everything we need to know for our ChatSession struct, including our first chain code. He will also inform other members about my public key and initial chain code.

So no need to initially broadcast a flood of initial keying messages. 

That means that to initially join a group, call add_subscribed_group() with the initial ChatSession struct all filled in. This same function can be re-used to restore a session from persistent store at next wallet session startup.

PR #1459, opened 2019-12-20
Finish off Private Messaging for general users

dbmcclain:
Adds general GUI functions for constructing private messages with binary data payloads, and decoding of incoming PaymentPayloadData::Data(Vec<u8>) into PrivateMessage types.


dbmcclain:
Todo: add a bunch more unit tests. So far just gets a clean compile.

PR #1463, opened 2019-12-24
Fix show balance panic.

vldm:
Because, we have broken implementation of Fr::from(hash:Hash) our utxo creation was incompatible with utxo processing, and sometimes produced utxo can fail to decrypt.

Also propogate error to upper level instead of panics.


dbmcclain (commented):
I would be surprised if this patch really fixes anything. If an Fr value deserializes to an invalid sized scalar, then deserializing first to hash won't improve that situation.

If this really does fix your problem, then the *real* problem lies down in the crypto deserialization routine, and by fixing here, you leave open the possibility that the error will re-occur somewhere else in the future.

PR #1493, opened 2020-02-17
Use nanosecond for timestamp, to prevent breaking timestamp during serialisation roundtrip.

vldm:


rtsisyk:
Replaced by https://github.com/stegos/stegos/pull/1494

PR #1506, opened 2020-05-22
Get going on the Github Actions workflow

joelreymont:


rust_toolchain
```diff
@@ -0,0 +1 @@
+nightly
```
vldm: Can you set toolchain version to some specific?

vldm (commented):
Maybe comment out clippy for now too? Or if ti possible mark it as failable?

rust_toolchain
joelreymont: Restored the old one.

vldm (approved):

